PR #14, opened 2023-03-03
fix unrecognized method blcok go routine and lack of second reply

yushimeng:


emiago (commented):
I added bigger fix in main branch. 
Thanks for submiting

PR #36, opened 2023-08-19
fix: Handle partial reads of messages from TCP socket

oystedal:
Hi,

First of all, thank you for making this library, I've found it very nice to work with for my personal SIP-based project so far.

The systems I work with are capable of producing large SDPs, which easily breaks the current assumption in `tcp.go` that requires all messages to be read in full. Often the headers of an INVITE would be parsed along with the first half of the SDP, and then there would be an attempt to parse the second half of the SDP as if it was a full message.

This commit introduces a mechanism to extract full SIP messages from a continuous stream of bytes. 

The current tests appear to pass with this change.

`HandlePartialMessages` has been added as a way to turn off this functionality for now in case it is needed to work around any issues that may arise. It is enabled by default.

### DIFF
--- a/parser/partial_parser.go
+++ b/parser/partial_parser.go
@@ -1,0 +1,155 @@
+package parser
+
+import (
+	"bytes"
+	"errors"
+	"strconv"
+)
+
+type partialMessageParserState int
+
+const (
+	Start   partialMessageParserState = 0
+	TopLine                           = 1
+	Headers                           = 2
+	Content                           = 3
+)
+
+var (
+	contentLengthHeader = bytes.NewBufferString("content-length:")
+)
+
+type PartialMessageParser struct {
+	state partialMessageParserState
+
+	inputBuffer_ *bytes.Buffer
+	currentLine_ *bytes.Buffer
+
+	hadContentLengthHeader bool
+	contentBytesLeft       int
+}
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func (p *PartialMessageParser) reset() {
+	p.state = Start
+	p.hadContentLengthHeader = false
+	p.contentBytesLeft = 0
+	p.inputBuffer_.Reset()
+	p.currentLine_.Reset()
+}
+
+func endsWithCRLF(buffer []byte) bool {
+	return len(buffer) >= 2 && buffer[len(buffer)-1] == '\n' && buffer[len(buffer)-2] == '\r'
+}
+
+func endsWithDoubleCRLF(buffer []byte) bool {
+	return len(buffer) >= 4 &&
+		buffer[len(buffer)-1] == '\n' &&
+		buffer[len(buffer)-2] == '\r' &&
+		buffer[len(buffer)-3] == '\n' &&
+		buffer[len(buffer)-4] == '\r'
+}
+
+func isContentLengthHeader(buffer []byte) bool {
+	if len(buffer) < contentLengthHeader.Len() {
+		return false
+	}
+
+	return bytes.EqualFold(contentLengthHeader.Bytes(), buffer[0:contentLengthHeader.Len()])
+}
+
+func (p *PartialMessageParser) Process(data []byte) (messages [][]byte, err error) {
+	i := 0
+
+	if p.currentLine_ == nil {
+		p.currentLine_ = bytes.NewBuffer(nil)
+	}
+
+	if p.inputBuffer_ == nil {
+		p.inputBuffer_ = bytes.NewBuffer(nil)
+	}
+
+	for i < len(data) {
+		dataByte := data[i]
+		switch p.state {
+		case Start:
+			{
+				if dataByte != '\r' && dataByte != '\n' {
+					p.state = TopLine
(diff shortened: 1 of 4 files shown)


emiago:
Hi @oystedal . First of all thnx for contributing and this cool approach of solving. I understand issue
Stream parsing I still did not found time to solve, and yes probably TCP will need different parser logic.  

I had a quick look and I see there are things for improve/change (correct me if I am wrong)
- No multi connection parsing (Single parser used), concurency, races
- double scanning could be reduced (parse once you get CR)
- callback func may be avoided, instead return once done

I would insist on some benchmarking and having performance near current parser. 
If you do not mind I may take this code and tests and do above, BUT as soon I find time. :)

oystedal:
Regarding your comments,

> No multi connection parsing 

Every TCPTransport will instantiate a new `PartialMessageParser`, so there is no data race or concurrency issue

> double scanning could be reduced (parse once you get CR)

The call to `bytes.HasSuffix(currentLine, []byte{'\r', '\n'})` can indeed be simplified to something that in theory should be more efficient.

> callback func may be avoided, instead return once done

True, one could return a slice of messages once we're done processing the current slice of bytes. That is probably a better interface.

> I would insist on some benchmarking

I agree. I did write a benchmark that I ended up not using. The reason is that the benchmark indicated that the code was extremely performant, so much so that I did not really believe the benchmark. I suspect the Go compiler is quite smart when it comes to avoiding making copies of slices where it's not needed, so one might need to trick the compiler a little to get a useful benchmark.

> If you do not mind I may take this code and tests and do above

Absolutely! Feel free to make changes to this PR as you see fit. You can also include portions of it in your own commits, all I ask is that you give me credits in the commit message in that case.

-------

There are some other things that could/should be addressed as well, either in this PR or in the future:

* Define a maximum message size (optionally adjustable at build time) so that we do not end up reading an unbounded amount of bytes from the TCP socket (possible DoS)
* Right now the `PartialMessageParser` includes its own parser for content-length, which duplicates the existing content-length parser. Additionally, the content-length header is parsed twice. It could be that the most elegant solution to this is to integrate the current parser more tightly into `PartialMessageParser` (that is, parse headers as they become available).
* The implementation is quite optimistic in its current form. If a message fails to parse we should ideally notify the UA to make it send a BYE and then close the connection. There probably needs to be an API to inform the application that the client has been forcefully disconnected.

emiago:
> Every TCPTransport will instantiate a new `PartialMessageParser`, so there is no data race or concurrency issue

I am not sure this is true anymore. TCPTransport is just a logic structure for creating and managing TCP connections.
Every new connection (Server or Client) starts reading. In that regard Parser must be thread/goroutine safe. 

For benchmarking I think we can reuse some existing. but anyway that will be easy part.

> Define a maximum message size (optionally adjustable at build time) so that we do not end up reading an unbounded amount of bytes from the TCP socket (possible DoS)
agree. Issue added #37 

> Right now the PartialMessageParser includes its own parser for content-length, which duplicates the existing content-length parser. Additionally, the content-length header is parsed twice. It could be that the most elegant solution to this is to integrate the current parser more tightly into PartialMessageParser (that is, parse headers as they become available).
Yes sure.
> The implementation is quite optimistic in its current form. If a message fails to parse we should ideally notify the UA to make it send a BYE and then close the connection. There probably needs to be an API to inform the application that the client has been forcefully disconnected.
Not sure for this. But something to also look
https://datatracker.ietf.org/doc/html/rfc3261#section-21.5.7

oystedal:
> I am not sure this is true anymore. TCPTransport is just a logic structure for creating and managing TCP connections.
Every new connection (Server or Client) starts reading. In that regard Parser must be thread/goroutine safe.

You are correct, I misunderstood the name. I have moved the partial parser to TCPConnection instead.

I have also updated the interface of the partial parser to return any full messages instead of the previous callback approach.

oystedal:
Added benchmark, used `bytes.Buffer`, and made other minor changes to make it run approx. 3 times faster compared to the initial implementation.

emiago:
@oystedal I appreciate changes. Looks better now. I need to recheck other parts as well.
Do you use some tools to generate this or what is your e2e testing?

oystedal:
No problem. I have no other changes planned at this time. Let me know if there's anything else you'd like me to change. 

For e2e testing I have used a server I have written using your framework. It communicates with an enterprise grade SIP device that I have access to. All the SIP messages and the SDP I have used in the test is generated by that device, with minor changes for privacy.

emiago:
@oystedal Stream parsing is now added on main branch and it is default for TCP and WS. 

Pls check  and if you do some testing to confirm behavior is correct. 
Once I get feedback I will see to push this new release. There is lot of fixes as well on main branch.

oystedal:
I did a few quick tests with your commit and it appears to work well.

```
require (
	github.com/emiago/sipgo v0.12.2-0.20230907182009-c9e30dc3b384
        // ...
)
```

Closing this PR as it is covered by https://github.com/emiago/sipgo/commit/c9e30dc3b3847d689da42241dfc7776e08b64dcd.

PR #41, opened 2023-09-09
Add fuzzer for SIP parser

oystedal:
Fuzzing input received from the network is a best practice for security and correctness. Go 1.18 added native support for fuzzing, see https://go.dev/doc/tutorial/fuzz and https://go.dev/security/fuzz.

The first commit introduces the fuzzer, which can be executed with `go test -fuzz=Fuzz`. For now, this starts off with a trivial SIP message. This can be expanded in the future by adding additional input strings as needed.

The subsequent commits are fixes to various issues found during execution of the fuzzer. Note that each of the files under `parser/testdata/fuzz` automatically become tests that are executed when running `go test`. This ensures that previously discovered crashes are not reintroduced with new changes.

I have executed the fuzzer for a little over two hours after fixing these issues without finding any additional issues.

### DIFF
--- a/parser/parse_uri.go
+++ b/parser/parse_uri.go
@@ -3,6 +3,7 @@ package parser
 import (
 	"fmt"
 	"strconv"
+	"strings"
 
 	"github.com/emiago/sipgo/sip"
 )
@@ -25,14 +26,12 @@ func ParseUri(uriStr string, uri *sip.Uri) (err error) {
 }
 
 func uriStateSIP(uri *sip.Uri, s string) (uriFSM, string, error) {
-	switch s[0] {
-	case 'S', 's':
-		if s[3] == 'S' || s[3] == 's' {
-			uri.Encrypted = true
-			return uriStateUser, s[5:], nil
-		}
+	if len(s) >= 4 && strings.EqualFold(s[:4], "sip:") {
 		return uriStateUser, s[4:], nil
-	default:
+	} else if len(s) >= 5 && strings.EqualFold(s[:5], "sips:") {
+		uri.Encrypted = true
+		return uriStateUser, s[5:], nil
+	} else {
 		return uriStateHost, s, nil
 	}
 }
--- a/parser/parse_uri.go
+++ b/parser/parse_uri.go
@@ -1,6 +1,7 @@
 package parser
 
 import (
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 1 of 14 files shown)


parser/parse_uri.go
```diff
@@ -27,11 +38,11 @@ func ParseUri(uriStr string, uri *sip.Uri) (err error) {
 func uriStateSIP(uri *sip.Uri, s string) (uriFSM, string, error) {
 	switch s[0] {
 	case 'S', 's':
-		if s[3] == 'S' || s[3] == 's' {
```
emiago, line 30: I think that maybe we should require sip/sips to be present, per comment above. This maybe would then simplify check

oystedal: Updated the code to check for `sip:` and `sips:`.

parser/testdata/fuzz/FuzzParseSipMessage/2c025c3ae292cb8b
```diff
@@ -0,0 +1,2 @@
+go test fuzz v1
```
emiago, line 1: I guess this was failing examples, but now they are fixed. 
Do we really need to hold on this?
Files are just very autogenerated and I guess they can be created with fuzz test every time (if not fixed already)

emiago:
@oystedal forgot to mention. Very cool, thnx for contributing. 
I will also run some benchs.

parser/parse_uri.go
emiago: ok this maybe now decreases performance. 
We need also to check upper Case. SIP, SIPS, that is why old implementation was looking both. 
I mostly wanted to remove default case, but keep this small check. 
This is how is done in some other implements, compromise performance and full check.

oystedal: > We need also to check upper Case. SIP, SIPS, that is why old implementation was looking both.

`strings.EqualFold` does case-insensitive matching. I added test cases to verify that we match both `sip:` and `SIP:`.

I don't think a huge amount of performance is lost here.

parser/testdata/fuzz/FuzzParseSipMessage/2c025c3ae292cb8b
oystedal: These files become test cases that help verify that the same kind of crash is not reintroduced into the parser later, so it's a free test case as such. If you undo one of my changes and run `go test` you'll see that the test fun fails.

These files are based on crashing cases that the fuzzer has found, they will not be generated just from new runs of the parser after the crash has been fixed.

I agree that the fuzzer ends up generating very quirky examples. Part of the reason is that it works hard to make the test cases as minimal as possible.

emiago:
merged. @oystedal

PR #42, opened 2023-09-10
fix: Log only received bytes instead of complete buffer

oystedal:
Logging of `string(b)` would log the entire buffer, regardless of message length. If a long message is read followed by a short message, the first part of the buffer would contain the short message, while the remainder of the buffer would be bytes from the long message that has not been overwritten by the short message.

Fix is to only log `string(b[:n])`, that is, the bytes of the buffer that were updated by `Read(b)`.


emiago:
ok cool. Merging

PR #45, opened 2023-09-22
Use default random source in `sip.RandStringBytesMask`

a12n:
A source created with [`rand.NewSource()`](https://pkg.go.dev/math/rand#NewSource) is not safe for concurrent use: "Unlike the default Source used by top-level functions, this source is not safe for concurrent use by multiple goroutines". There may be multiple goroutines preparing SIP requests and doing `sip.GenerateTagN()`. So, it's unsafe to use global `var src = rand.NewSource()`.

That's how it look with race detector:
```
WARNING: DATA RACE
Read at 0x00c0000e6000 by goroutine 11:
  math/rand.(*rngSource).Uint64()
      /usr/local/go/src/math/rand/rng.go:239 +0x30
  math/rand.(*rngSource).Int63()
      /usr/local/go/src/math/rand/rng.go:234 +0x1cd
  github.com/emiago/sipgo/sip.RandStringBytesMask()
      /root/go/pkg/mod/github.com/emiago/sipgo@v0.12.1/sip/utils.go:53 +0x58
  github.com/emiago/sipgo/sip.GenerateTagN()
      /root/go/pkg/mod/github.com/emiago/sipgo@v0.12.1/sip/sip.go:50 +0x947
…
Previous write at 0x00c0000e6000 by goroutine 12:
  math/rand.(*rngSource).Uint64()
      /usr/local/go/src/math/rand/rng.go:239 +0x44
  math/rand.(*rngSource).Int63()
      /usr/local/go/src/math/rand/rng.go:234 +0x1cd
  github.com/emiago/sipgo/sip.RandStringBytesMask()
      /root/go/pkg/mod/github.com/emiago/sipgo@v0.12.1/sip/utils.go:53 +0x58
  github.com/emiago/sipgo/sip.GenerateTagN()
      /root/go/pkg/mod/github.com/emiago/sipgo@v0.12.1/sip/sip.go:50 +0x947
…
```

Is there a reason to not use the default source? As the neighbouring `sip.RandString()` does.

### DIFF
--- a/sip/utils.go
+++ b/sip/utils.go
@@ -44,15 +44,13 @@ func RandString(n int) string {
 	return string(output)
 }
 
-var src = rand.NewSource(time.Now().UnixNano())
-
 // https://stackoverflow.com/questions/22892120/how-to-generate-a-random-string-of-a-fixed-length-in-go
 func RandStringBytesMask(sb *strings.Builder, n int) string {
 	sb.Grow(n)
-	// A src.Int63() generates 63 random bits, enough for letterIdxMax characters!
-	for i, cache, remain := n-1, src.Int63(), letterIdxMax; i >= 0; {
+	// A rand.Int63() generates 63 random bits, enough for letterIdxMax characters!
+	for i, cache, remain := n-1, rand.Int63(), letterIdxMax; i >= 0; {
 		if remain == 0 {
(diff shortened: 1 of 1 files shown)


emiago:
Great finding, did not saw that. I checked is there some big performance penalty as well.

PR #49, opened 2023-10-04
Call TransactionRequest with context

Maratk1n:
Subject allows to specify timeouts, e.g.:
```
ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
defer cancel()

tx, err := client.TransactionRequest(ctx, req, opts...)
```
By default, connection establishment may take 2-3 minutes depending on the transport.

https://pkg.go.dev/net#Dialer
> // With or without a timeout, the operating system may impose
> // its own earlier timeout. For instance, TCP timeouts are
> // often around 3 minutes.
> Timeout [time](https://pkg.go.dev/time).[Duration](https://pkg.go.dev/time#Duration)


emiago:
This is something I was probably considering for some time.
Thnx for contributing. Needs rebase but I will check what is effect also on performance.
@Maratk1n I also considering having ctx for server side transactions. Still not sure what is best approach.

Maratk1n:
I rebased branch and resolved conflicts.

emiago:
Hey @Maratk1n I approve this, just for easier history check,  if not big deal to squash this changes.

emiago:
We also have to note this is not standard req/resp pattern as still control is moved after setup conn.
This ctx has affect on conection setup (which are not closed after) but bigger use is for DNS (Loadbalancing via DNS)
This are delays mostyl comming.

so we need to change `defaultResolver` as well here. It can be another ticket well. It uses behind ctx
`DefaultResolver.internetAddrList(context.Background(), afnet, address)`

Maratk1n:
Hi, @emiago . No problem, I've squashed commits.

PR #65, opened 2024-01-02
Correctly reload dialog state

dennwc:
Currently the dialog state is loaded once and checked in the loop. This change properly reloads the current state on each loop iteration.

### DIFF
--- a/dialog_server.go
+++ b/dialog_server.go
@@ -208,6 +208,7 @@ func (s *DialogServerSession) Bye(ctx context.Context) error {
 	// until it has received an ACK for its 2xx response or until the server
 	// transaction times out.
 	for {
+		state = s.state.Load()
 		if sip.DialogState(state) < sip.DialogStateConfirmed {
 			select {
 			case <-s.inviteTx.Done():


dialog_server.go
```diff
@@ -208,7 +207,7 @@ func (s *DialogServerSession) Bye(ctx context.Context) error {
 	// until it has received an ACK for its 2xx response or until the server
 	// transaction times out.
 	for {
-		if sip.DialogState(state) < sip.DialogStateConfirmed {
+		if sip.DialogState(s.state.Load()) < sip.DialogStateConfirmed {
```
emiago, line 210: Lets keep initialization seperate. Easier debuging. 
So just have `state` var reinitialized.

Good catch

dialog_server.go
```diff
@@ -185,9 +185,8 @@ func (s *DialogServerSession) WriteResponse(res *sip.Response) error {
 }
 
 func (s *DialogServerSession) Bye(ctx context.Context) error {
-	state := s.state.Load()
```
emiago, line 188: pls return initialization. Same as below

dialog_server.go
dennwc: Done!

emiago:
@dennwc Thnx for fixing. Merging...

PR #67, opened 2024-01-05
fix: correctly parse addresses with user tags and header tags

InnovativeUS:
Correct the parsing of addresses with both user and header tags.


emiago:
Merging, seems it was not issue if there is only one params present, but both it is issue.

PR #68, opened 2024-01-17
Handle 407 Proxy-Authenticate header with digest invite

AnshulMalik:


emiago:
Hi @AnshulMalik thnx for contributing. 
I think after merging this will need some refactor. 
Maybe I see wrong but only difference is status code and header that is included.

PR #74, opened 2024-02-25
feat: Add OnBye handler to Client

### LINKED ISSUE #73: API for handling Bye as a client
I don't believe anything exists today to handle Bye as a client. Would it be ok if I added this API? 

```
func (c *Client) OnBye(handler RequestHandler) {
```

Sean-Der:
Resolves #73

### DIFF
--- a/client.go
+++ b/client.go
@@ -21,6 +21,9 @@ type Client struct {
 	port  int
 	rport bool
 	log   zerolog.Logger
+
+	// requestHandlers map of all registered request handlers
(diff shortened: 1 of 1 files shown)


emiago:
sorry but this is not how this will work. 

Client and Server are splitted intentionally to provide seperate control. 

As long you use same `UA` it can act as both sender and receiver, just that you use Client and Server structure to seperate 
sending and receiving side.

We can get in meeting if needed ;)

PR #78, opened 2024-03-13
fix: use invite response route headers for bye

AnshulMalik:
`Record-Route` header  from invite response  should be converted to `Route` header in the bye request.

### DIFF
--- a/sip/request.go
+++ b/sip/request.go
@@ -347,10 +347,10 @@ func NewByeRequestUAC(inviteRequest *Request, inviteResponse *Response, body []b
 	if len(inviteRequest.GetHeaders("Route")) > 0 {
 		CopyHeaders("Route", inviteRequest, byeRequest)
 	} else {
-		hdrs := inviteResponse.GetHeaders("Record-Route")
-		for i := len(hdrs) - 1; i >= 0; i-- {
-			h := hdrs[i].headerClone()
-			byeRequest.AppendHeader(h)
+		recordRoute := inviteResponse.RecordRoute()
+		for recordRoute != nil {
+			byeRequest.AppendHeader(&RouteHeader{Address: recordRoute.Address})
+			recordRoute = recordRoute.Next
 		}
 	}
 


sip/headers.go
```diff
@@ -920,6 +920,7 @@ func (h *ContentTypeHeader) headerClone() Header { return h }
 // RouteHeader  is Route header representation.
 type RouteHeader struct {
 	Address Uri
+	Next    *RouteHeader
```
emiago, line 923: Sorry but I have removed this linked list parsing for comma seperated values. 
They add more complexity, so I am not sure have you rebased on latest?

I would like to add fix for BYE

AnshulMalik: Thanks, fixed that now.

sip/request.go
```diff
@@ -347,10 +347,10 @@ func NewByeRequestUAC(inviteRequest *Request, inviteResponse *Response, body []b
 	if len(inviteRequest.GetHeaders("Route")) > 0 {
 		CopyHeaders("Route", inviteRequest, byeRequest)
 	} else {
-		hdrs := inviteResponse.GetHeaders("Record-Route")
```
emiago, line 350: I have not noticed but this is break change.
RRs can be multiple.  Can you tell me what RFC part you are trying to fix. 
I know lib is not fully handling route headers

AnshulMalik: https://datatracker.ietf.org/doc/html/rfc3261#section-16.12.1.2

AnshulMalik: There is even fine grained rule of how to do this properly, that I didn't do here.

emiago: Hi @AnshulMalik . Here is the thing. This function I already have a deprecated status, and you may see it is dedicated to UAC, but not UAS.
Now your request is moved on `DialogServerSession` logic. So pls use dialog for handling. 
I have not fully test this, but pls confirm.


All proxy related stuff that has to deal with route and record route headers will be first developed as part of 
wsip (https://github.com/emiago/wsip). Currently I am only preparing this project, but you can check the source.

AnshulMalik: Okay, you mean to say that we'll move some things from sipgo to wsip and use those?
Because UAC and UAS in sipgo need to understand and act on the record-route headers set by proxies in the middle. Which is currently broken in dialogsession and dialogclient.

emiago: I have also fixed record-route within dialog. 
for wiresip, yes I am thinking to invest more there for  handling this type of behavior as proxy, but if you are building client or server with dialog, now you should not have issue speaking with proxies.
Record-Route and Route headers should be covered.

So SIP has different behavior: 
- UAC 
- UAS
- Proxy

AnshulMalik: This is what I was trying to explain earlier that current implementation is not considerate to intermediate proxies.
https://datatracker.ietf.org/doc/html/rfc3261#section-16.12.1.2

If there are multiple proxies in between, our bye requests from UAC don't read record-route from the invite response.

emiago: Hi @AnshulMalik can you debug it and see why does not this apply or we are doing something wrong. 

Route headers are read on transport layer. So request should be sent. I did not test this too much, so probably something is wrong

Here is the line of code
https://github.com/emiago/sipgo/blob/main/dialog_server.go#L319

emiago: Also for client 
https://github.com/emiago/sipgo/blob/main/dialog_client.go#L282

AnshulMalik: Oh, didn't see this client change, might be fixed now, I'll check.

PR #87, opened 2024-05-09
Change AppendHeaderAfter to append after the provided header name pas…

KingDanx:
…sed rather than before.

### DIFF
--- a/sip/headers.go
+++ b/sip/headers.go
@@ -162,7 +162,7 @@ func (hs *headers) AppendHeaderAfter(header Header, name string) {
 		return
 	}
 
-	if ind+1 > len(hs.headerOrder) {
+	if ind+1 == len(hs.headerOrder) {
 		hs.AppendHeader(header)
 		return
 	}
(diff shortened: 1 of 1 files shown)


KingDanx:
AppendHeaderAfter() was prepending the passed header to the provided header rather than appending as the function is indicating.

emiago:
@KingDanx thnx for fix, but would it not just increasing `ind` by 1 (after checks), kept most of code and a little bit easier on reading.

KingDanx:
@emiago I'm fairly new to Go so please forgive if it's hard to read.  

If ```newOrder[ind+1] = header``` is the only thing that is changed, it will be overwritten by the next 
```copy(newOrder[ind+1:], hs.headerOrder[ind:])``` call.

In order to populate the remainder of the headers into ```newOrder``` after the new header is added at ```ind+1```, I had to go to the next index: ```ind+2```.

In order to avoid going out of bounds when the name string header is the last index I added the check call ```AppendHeader(header)``` and ```return```. However I now see that the logic on that check is incorrect. it should be an equivalency test.
```Go
	if ind+1 == len(hs.headerOrder) {
		hs.AppendHeader(header)
		return
	}
```

I'm sure anyone more experienced with the language can express this in an easier to read manner. Let me know what you think.

emiago:
@KingDanx Only fix that one line. I will merget after. Thanks for contribution

PR #91, opened 2024-05-21
set min to 1.20 for win 7 compliant

yangjuncode:
we use sipgo in a project which deploy under windows 7, the go 1.20 is last version support windows 7, so please lower one version for win 7 compliant， sipgo don't use new feature in newer version as far as i know.

### DIFF
--- a/example/proxysip/docker/proxy/Dockerfile
+++ b/example/proxysip/docker/proxy/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.17
+FROM golang:1.20
 
 WORKDIR /go/src/app
 COPY . .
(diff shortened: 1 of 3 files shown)


emiago:
There are tests, I think using this new keywoards (min,max) as for compilation. Could be removed yes. 
Anyway I can not guarantee this will stay on this version even if I downgrade. `sipgox` is probably even more dependent. 

What about forking?

yangjuncode:
i don't know the form new keywoards (min,max)  in go.mod, any doc url for it?
of cause, if someday sipgo use new feature > 1.20, my project have to stay on the version which support it.
as for now, 1.20 is ok.
yeah, i have to fork, but it would be good if  upstream can support 1.20 for now, since sipgo is in fast development state.
anyway thanks for your work!

emiago:
Moving this to issue, as more things need to be reviewed.

emiago:
fixed

Issue #95, opened 2024-06-14
Could you provide a client-side example, with TCP long connection


duiniuluantanqin:
Could you provide me with a client-side example to achieve the following functionality?

1. The client actively connects to the server (another remote host) via the TCP protocol. 
2. Once the connection is established successfully, it sends a REGISTER message proactively and maintains this TCP long connection. 
3. It can then receive INVITE, BYE, MESSAGE, and other messages sent actively by the remote party and respond accordingly.

emiago:
hi @duiniuluantanqin . There is some example for register in repo, but for you problem
you only need to setup this 
https://github.com/emiago/sipgo/blob/main/sip/transport.go#L16

Anyway by default sipgo is keeping connection open as long they are referenced. in other words as long you have active transactions (NOT terminated). 

On other hand you have all that covered on `sipgox.Phone` with `Answer`.  So checkout https://github.com/emiago/sipgox repo

You only have one line to call and voilaa. HINT: it works only for single CALL.

Feel free to reopen issue if you need more details

duiniuluantanqin:
Thank you very much for your response.
However, I have another question regarding the third point mentioned above. I want to implement a Client (A) that connects to a Server (B) via the TCP protocol.
By calling `ClientRequestConnection`, I can obtain a socket connection, such as `conn Connection`.
Now, I would like to set some callback events for `conn`, such as `OnInvite`, `OnBye`, `OnMessage`. How can this be achieved?

PR #96, opened 2024-06-17
feat: Add request handlers for client

duiniuluantanqin:
for https://github.com/emiago/sipgo/issues/95

This is just one approach. If you find it feasible, I will continue to refine it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -23,6 +23,9 @@ type Client struct {
 	port  int
 	rport bool
 	log   zerolog.Logger
+
+	requestHandlers map[sip.RequestMethod]RequestHandler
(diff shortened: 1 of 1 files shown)


emiago:
This is already provided. Maybe I need to make this more clear in docs. You have `Server` for this. 
client and server are just **handles** (wrappers) for transport layer and transport layer is created by **user agent**, specifically by constructor
`NewUA()` 

So as long you are sharing same `UserAgent` accross client and server handles, you are utilizing same tranport layer. 
In your case, just attach your client `ua` to `NewServer(` and you are good to go.

duiniuluantanqin:
Got it. Thank you .

PR #98, opened 2024-06-18
fix(sip): default the SIP URI port to 5060 if it is not provided

InnovativeUS:
Twilio does not provide the port of the URI address. This pull request adds a default port of 5060 to the URI so that we can send the response back to the right place.


emiago:
Hi @InnovativeUS lets consider some better fix.  "ip:0" is valid for back GO transport. It could be that is  not used, but it adds friction. 

The more correct way is also to include transport as for SIP TLS , default is 5061. 
In `sip` package there are constants defined for this.

Add some context to issue. Transport layer handles empty port in uri, so I am not sure where is exactly issue?

PR #99, opened 2024-06-18
fix: honor lr in record route for dialog clients

InnovativeUS:
The LR in a record route means that the route should be added to the client request. This fix honors the specification.


emiago:
LGTM

PR #100, opened 2024-06-19
fix: ACK request does not convert the Record-Route headers into Route headers

hateeyan:
UAC ACK request should convert the Record-Route headers into Route headers.

https://datatracker.ietf.org/doc/html/rfc2543#section-6.29

> The calling user agent client copies the Record-Route header into a
   Route header field of subsequent requests within the same call leg,
   reversing the order of requests


emiago:
@hateeyan this function is for me deprecated and I will now remove from API. 

I will port your fix on better place. Thanks for noticing this. 

Will reference you in commit

emiago:
Actually I will merge this, and then do later refactor. Thanks

PR #111, opened 2024-07-30
bug: uri.HostPort() unset port

yallxe:
Port zero returned by this function caused some internal bugs, for example when using DialogClientSession's Ack() function with Record-Route header specified in the response.

This PR should fix this.

### DIFF
--- a/sip/uri.go
+++ b/sip/uri.go
@@ -131,6 +131,9 @@ func (uri *Uri) Addr() string {
 
 // HostPort represents host:port part
 func (uri *Uri) HostPort() string {
+	if uri.Port == 0 {
+		return uri.Host
+	}
 	p := strconv.Itoa(uri.Port)
 	return uri.Host + ":" + p
 }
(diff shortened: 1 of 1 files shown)


emiago:
I think correct behavior is to return actually host:0. 
This PR was asked before, so I am now curious, on which line this gets called.

yallxe:
Here this function is called under the condition that Record-Route header is present in the invite response. 
 
https://github.com/emiago/sipgo/blob/e02d4df836c22c7e40d9caa415986545a8b41cdd/dialog_client.go#L305-L310

emiago:
ok, now it is clear what should be fixed

emiago:
@smlsr should be fixed on transport layer. Pls retest latest main

yallxe:
@emiago It seems that this doesn't fix the issue. With sip.SIPDebug enabled, I can see that these lines in my logs:
```
Aug  2 16:25:08.444688 DBG UDP write to 192.168.1.108:59928 -> 123.123.123.123:0:
ACK sip:1111@123.123.123.123:5060 SIP/2.0
...
```
And the ACK is still being sent to port zero.

emiago:
@yallxe now it should be UDP fixed

PR #113, opened 2024-08-02
export SetIP function

dwdcth:
On my cloud server, the automatically obtained IP is a local LAN IP rather than a public IP, causing message delivery failures to cloud server form devices. I need a method to expose this setIP functionality for manual IP configuration.


emiago:
@dwdcth this will be probably deprecated. Tell me why are you not using 
WithClientHostname(). 
This is only where is this used as a backup.

dwdcth:
Ok, Thank you.

PR #114, opened 2024-08-02
Added tel: URI support

IvanRibakov:
This PR adds simplified support for the [tel: URIs](https://datatracker.ietf.org/doc/html/rfc3966#section-3).

No validation on the telephone number is performed.

Attempted to preserve existing URI type without diving into different interfaces for different URI types.

### DIFF
--- a/sip/parse_uri.go
+++ b/sip/parse_uri.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-	"unicode"
 )
 
 type uriFSM func(uri *Uri, s string) (uriFSM, string, error)
@@ -48,10 +47,6 @@ func uriStateScheme(uri *Uri, s string) (uriFSM, string, error) {
 
 	uri.Scheme = strings.ToLower(s[:colInd])
 	s = s[colInd+1:]
-
-	if err := validateScheme(uri.Scheme); err != nil {
-		return nil, "", err
-	}
 
 	if uri.Scheme == "tel" {
 		return uriTelNumber, s, nil
@@ -177,24 +172,6 @@ func uriStateHeaders(uri *Uri, s string) (uriFSM, string, error) {
 	uri.Headers = NewParams()
 	_, err = UnmarshalParams(s, '&', 0, uri.Headers)
 	return nil, s, err
-}
-
-// validateScheme performs basic scheme validation to prevent cases where port delimited 
-// (or other naturally occuring colon) is mistakenly used as a scheme delimiter. 
-// This is NOT a fool-proof validation and URIs may still be incorrectly parsed 
-// unless more parsing validation effort is made.
-//
-// scheme        = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
-func validateScheme(scheme string) error {
-	if len(scheme) == 0 {
-		return errors.New("no scheme found")
-	}
-	for _, c := range scheme {
-		if !unicode.IsLetter(c) && !unicode.IsDigit(c) && c != '+' && c != '-' && c != '.' {
-			return fmt.Errorf("invalid scheme: %q is not allowed", c)
-		}
-	}
-	return nil
 }
 
 func uriTelNumber(uri *Uri, s string) (uriFSM, string, error) {
--- a/sip/parse_uri.go
+++ b/sip/parse_uri.go
@@ -16,7 +16,8 @@ func ParseUri(uriStr string, uri *Uri) (err error) {
 	if len(uriStr) == 0 {
 		return errors.New("Empty URI")
 	}
-	state := uriStateSIP
+
+	state := uriStateStart
 	str := uriStr
 	for state != nil {
 		state, str, err = state(uri, str)
@@ -27,16 +28,7 @@ func ParseUri(uriStr string, uri *Uri) (err error) {
 	return
 }
 
-func uriStateSIP(uri *Uri, s string) (uriFSM, string, error) {
-	if len(s) >= 4 && strings.EqualFold(s[:4], "sip:") {
-		return uriStateScheme, s[4:], nil
-	}
-
-	if len(s) >= 5 && strings.EqualFold(s[:5], "sips:") {
-		uri.Encrypted = true
-		return uriStateScheme, s[5:], nil
-	}
-
+func uriStateStart(uri *Uri, s string) (uriFSM, string, error) {
 	if s == "*" {
 		// Normally this goes under url path, but we set on host
 		uri.Host = "*"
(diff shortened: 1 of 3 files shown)


emiago:
Hi @IvanRibakov . First of all good work. Inteface I would reject ;). 
Anyway I need to justify reason on complicating this, URI is URI. I could see point having SCHEME types, to go in more generic way. I believe `tel` is still URI per RFC.
Of course we in SIP need better parsed URIs for access, but stdlib can also do the job.

So what I suggest, keep SCHEME detection, fix. (No need for constants SCHEME_ etc..., it can be http even) . User=Telephone number should be left. For rest unless fix needed, parsing should be completed.

Now this uris are not IP routable, so it must be some other way you are using them. This is what interest me. 

Anyway calling Endpoint, Addr, Hostport should be **Caller** Aware thing, and it must expect weird results if using on different type of URIs.

IvanRibakov:
> Now this uris are not IP routable, so it must be some other way you are using them. This is what interest me.

We have a SIP B2BUA that we're considering porting to Go. TEL URIs come up when dealing with:

1. Participant identification via `P-Asserted-Identity` headers which could contain both SIP and TEL URIs. This is quite important as once TEL URI was used to identify participant, it is then likely to be used in a lot of other places. For example - HTTP-enabled Location Delivery (https://datatracker.ietf.org/doc/html/rfc6155)
2. Call routing - calls to TEL URIs are routed via dedicated gateways

-------

> Inteface I would reject ;).

May I ask why (out of curiosity and to better understand the validation criteria applied to the PRs for the future)?

IvanRibakov:
> So what I suggest, keep SCHEME detection, fix. (No need for constants SCHEME_ etc..., it can be http even) .

Are you saying that you want want URI struct to be able to represent any kind of scheme without explicitly listing supported schemes and checking them during parsing?

In that case you may want to reconsider the decision to support [invalid URIs that don't contain the scheme](https://github.com/emiago/sipgo/blob/main/sip/parse_uri_test.go#L127). Otherwise the parsing logic is ambiguous when it comes to identifying the scheme. According to the [RFC](https://www.rfc-editor.org/rfc/rfc3986#section-3) only the "URI-reference" may not contain the scheme, but in that case ir must start with a double forward slash (see `relative-ref` below):

```
   URI           = scheme ":" hier-part [ "?" query ] [ "#" fragment ]

   hier-part     = "//" authority path-abempty
                 / path-absolute
                 / path-rootless
                 / path-empty

   URI-reference = URI / relative-ref

   absolute-URI  = scheme ":" hier-part [ "?" query ]

   relative-ref  = relative-part [ "?" query ] [ "#" fragment ]

   relative-part = "//" authority path-abempty
                 / path-absolute
                 / path-noscheme
                 / path-empty
```

IvanRibakov:
@emiago I made some changes to the PR according to my understanding of the feedback that you provided. Could you please take a look?

----

> Anyway I need to justify reason on complicating this, URI is URI.
> ...
> Of course we in SIP need better parsed URIs for access, but stdlib can also do the job.
> ...
> Now this uris are not IP routable, so it must be some other way you are using them.

I got impression from above that there is some reluctance to complicate URI parsing/storage/serialisation (presumably for performance reasons) since for the purposes of this library it is enough to support SIP URIs. Have you considered adding a pluging/extension mechanism for URIs in the same way as support for custom headers can be added via `DefaultHeadersParser()`? This way library users may add their own parser to support schemes other than SIP/SIPS? Ideally this solution would allow defining new structs to better represent other schemes and to isolate the serialisation and any other specific logic.

The reason I'm asking is because next I was going to create another PR to add support for URNs (still technically a URI). And if I were to proceed with the current trajectory of putting everything into a single type/parser, it will eventually lead to a solution that is hard to maintain.

In case you're wondering, [URNs are often used](https://datatracker.ietf.org/doc/html/rfc5031) as request URIs in emergency calls: `urn:service:sos.police` etc.

emiago:
Hi @IvanRibakov yes, performance here is key,  and validation should be minimal.  Benchmarks would be needed to verify.
Anyway, I see where you are going, and I think you already answered my concerns. 
I think this all are Great stuff to add, but only thing I would change here is having scheme for now if possible. (this would remove sip prefix checks in parsing yes)
Encrypted value is also something I am thinking like overhead added by me. This can be just post scheme check without big overhead (or cache on first read).

In other words I would like to see no special cases when deserializing and serializing based on scheme type. I understand having special fields looks better but this is then different thing.

The fact that this uris are not routable by SIP is something that you are discarding? And I think this even is mention in RFC, in order to have any usage of them, you have to make them SIP uri routable for SIP uas to even act on them. `SetDestination` is override NOT set.
So if this are used in routing then this will not be valid.

Anyway here how I currently see this, uris should be parsed (with your scheme fix), but not completely and not fully validated.
So for now I would consider post Parsing and validation as good contribution. `Lazy parsing` could be also something to consider, but I think it could be overhead for performance, and you can reduce this on caller side anyway.
What I am saying, unless it is buggy, all values from this Uris will find their place in current SIP uri structure? Now reading and additional parsing I am fine having but as helper functions, to keep smaller interface as possible on URI itself. So you can have this conversions implemented in some sense or extend URI.
uri <-> teluri
uri <-> urn 

Parsing and validation on parser level, it is bigger topic, and this is could be also extended, but lets have this through Issue discussed more, and not with PRs.

Also thanks for this noticing 
> https://github.com/emiago/sipgo/blob/cdeb0ceb3a958c1a44cc56b1e8ee785d6d69bb7b/sip/parse_uri_test.go#L79-L86

I think this was validated before as error.

IvanRibakov:
> only thing I would change here is having scheme for now if possible. (this would remove sip prefix checks in parsing yes)

Done in https://github.com/emiago/sipgo/pull/114/commits/0bc4daadb6ea742f102fb3b8b6e36c0c9b545bf7

> I would like to see no special cases when deserializing and serializing based on scheme type.

Not 100% sure what you mean by this. Are you saying that you don't want to see patterns like "if scheme is X then parse/serialize Y", for example:
- https://github.com/emiago/sipgo/pull/114/files#diff-767b81d8a15044e2632e0dc13a630d900eaffcc885434a65761586f314b08862R51-R53
- https://github.com/emiago/sipgo/pull/114/files#diff-4fca781cf75b4002030eae12cb22c4c944555600573b3043f21cd22f6ca37b2eL64-R71

emiago:
@IvanRibakov correct. So not to go back and forth, do you mind I take from here, as fix should be minimal. I will add only this scheme fix, and some refactoring. You should then be able to build up URI conversions. Some proposal could be 
TelUri.FromUri() TelUri.ToUri().

IvanRibakov:
> So not to go back and forth, do you mind I take from here

Fine by me.

> I will add only this scheme fix, and some refactoring.

I guess then I should wait for you to complete this before creating new PR for URN support? Or will you consider improving URN support yourself as well?

emiago:
hey @IvanRibakov I added now more general version of uri, but still keeping old performance

Now as supporting this additional uri types. Conversions must be in place. 
So you as caller need to identify scheme and call proper uri conversion. 

Maybe even this needs seperate package. 
`sipconv` or uriconv. 

Anyway I hope this unblocks you in your fork or lets create this as issue, and maybe more people start asking this.

Issue #119, opened 2024-08-12
DNS Resolution Error in case of WSS based SIP Transaction


p-akshay:
Hello,

I am working on a system where my SIP Server is on AWS.
in my setup, I have "A Record" mapping of "ims.xxxx.in" as the domain-name to "65.2.x.x" IP address. 

On most networks, I can use my SIP Server from my client via forwarding the packet to "ims.xxxx.in" and using the WSS as the transport.

However, there are few networks, where this server is not accessible as the SIP request to server timeouts.
Upon further investigation, what I found is as following:

1. If we do `net.LookupIP("ims.xxxx.in")` on a problem network and print the results then, we receive the following: 
`IP address: 64:ff9b::4102:xxxx
IP address: 65.2.x.x`

2. Looking at sipgo transport-related code, I can see that the function "resolveAddrIP()" used for DNS lookup, returns the 1st IP address found for the given host.
3. In my scenario, since the 1st IP address is of type IPv6 (actually IPv4 represented in IPv6 Format), and it is not directly mapped to the given host, I think all the SIP transactions are timing out.

Is there a way to solve this issue? like can we do some kind of cascading type IP address pickup here?
Or can we put some kind of restrictions for looking for IPv4 addresses only?

Any suggestion will be helpful.

Thanks

emiago:
Hi @p-akshay, currently not solvable. We could btw extend resolver behavior. Unfortunately go resolver is not much configurable, so it would require wrapper.

emiago:
Anyway this is something that will be addressed more. I suggest going with some temporarly fix. It can be option on transport layer. Give me some time to add it.
For example DNSAddrFilter(hostname string, []net.IP) []net.IP

p-akshay:
Hi @emiago ,

Ok...noted...

I am using a temporary workaround where I check if the resolved IP address is IPv4 or not and use the correct one.

Thanks for the explanation.
If you want, I can close the Issue.

Regards

emiago:
@p-akshay this is now forced to be IPV4. I would consider your issue is fixed, but we still need better way to solve.

p-akshay:
Ok Thanks

will test and confirm

PR #124, opened 2024-08-20
Prioritize IPv4 during DNS resolution

gpi-eaglys:
This modification prioritizes IPv4 address during DNS resolution.
It solves issues when IPv6 is returned, but not routable. 
See also  https://github.com/emiago/sipgo/issues/119.

### DIFF
--- a/sip/transport_layer.go
+++ b/sip/transport_layer.go
@@ -468,6 +468,12 @@ func (l *TransportLayer) resolveAddrIP(ctx context.Context, hostname string, add
 		return fmt.Errorf("lookup ip addr did not return any ip addr")
 	}
 
+	for _, ip := range ips {
+		if len(ip.IP) == net.IPv4len {
+			addr.IP = ip.IP
+			return nil
+		}
+	}
 	addr.IP = ips[0].IP
 	return nil
 }


emiago:
This change is hardcoding behavior. That is why it needs to be more caller handled.

gpi-eaglys:
The current non-deterministic approach is also hard-coded. It picks the first address from a random list `ips[0].IP`. But why not the last one? 


I agree that the PR is not a proper solution, but keeps many users happy who cannot handle IPv6, at minimum cost.

The proper solution would probably be to specify an interface for `sipgo.WithUserAgentDNSResolver()` and provide a sensible default, something like `IPv4FirstDnsResolver`. 
Why an interface? 
* `sipgo` does not need all the 20 member functions of the `net.Resolver` struct, it uses only 3
* what `sipgo` actually needs is a single address from the resolver.  It should not have any preferences which element of the list it picks. 
 
I agree that the library should be neutral wrt DNS resolution strategies. 
But I don't think that a randomized selection of either IPv6 or IPv4 is a sensible default. Making IPv4 the default would save many man-hours.

emiago:
@gpi-eaglys give me a week, if not I will merge this PR. We need to sort this out before 1.0. Yes interfacing dns is not good.

gpi-eaglys:
Thank you for the prompt response. Let us know if we can help with something.

PR #126, opened 2024-09-02
feat:add user-agent in headers

muzi-long:
add user-agent in headers


emiago:
I hope closing is due to fact it can be read just as string ;)

PR #134, opened 2024-09-26
Add Re-INVITE for DialogSession

wavexiang:


emiago:
@wavexiang al more extra dialog func will be moved to diago. Please next time open issue. 
Re Invites is already implemented as part of diago project. 
SIPGO will not have any more feature request as goal is to move to stable version.

PR #139, opened 2024-10-15
Return pointer from NewConnectionPool

dennwc:
Returning by value triggers static analyzers that detect that the mutex inside the connection pool is copied.


emiago:
nice

PR #140, opened 2024-10-18
add headers uac features

SilvaMendes:
Add to the uac response the facilities that the client offers as an endpoint.
pullrequest

### DIFF
--- a/sip/response.go
+++ b/sip/response.go
@@ -218,6 +218,13 @@ func NewResponseFromRequest(
 		res.AppendHeader(h.headerClone())
 	}
 
+	// RFC 3261 - 20.5
+	CopyHeaders("Allow", req, res)
+	// RFC 3261 - 20.37
+	CopyHeaders("Supported", req, res)
+	// RFC 3265 -  3.3.7
+	CopyHeaders("Allow-Events", req, res)
+
 	// 8.2.6.2 Headers and Tags
 	// the response (with the exception of the 100 (Trying) response, in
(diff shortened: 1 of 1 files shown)


emiago:
@SilvaMendes Thanks for contribution. Please share how you are using this function? 
Please try to avoid PR before issue discussion. I hate to reject stuff.
 I am maybe missing some more context here, but those headers do not belong here.

emiago:
@SilvaMendes Thanks for contribution. Please open issue first. I hate to reject stuff. 
I see no point here this adding. Those headers are optional. What problem you have, why you think changing this makes more sense?

SilvaMendes:
To maintain the compatibility of the resources that the endpoint offers, taking into account that in my case I am not using opensips with the b2c_logic module, so it does not break the initial tm request.

emiago:
I understand what this headers mean, but having that implicit behavior is questionable. Normally you as server or as client should build this up. I think this things need to go either on client or server. 
sipgo needs to be minimalistic, this is API to provide minimal (MUST / correct) response. All optional must be treated as optional. I see more having this on diago project.
FYI: this things I have build manually either as part of response or as custom prebuilt request. 

I think this could be worked more on client or server, but still leave caller to wrapup this behavior. 
Changing this to issue. To give more context, please take time to explain on issue what happens if this headers are not present in your case.

SilvaMendes:
By default it is up to the client or the server to decide whether to keep the backing fields.

PR #151, opened 2024-11-04
Proposal for fixing ACK destination on INVITE

DominikRidjic:
* ACK is in some scenarios sent to the wrong destination. Fixed by first reversing record routes and then using the first entry as destination.
* In our tests we saw this causing calls using UDP to fail, TCP worked.
* Example of failure before change: https://app.dev.sipfront.com/kiosk/bf6c0b85-c1bd-497f-b8b3-71772e935f37/94626162-9af0-11ef-adf3-b36e6b5447b2
* Example of behaviour after change: https://app.dev.sipfront.com/kiosk/bf6c0b85-c1bd-497f-b8b3-71772e935f37/f5dfd402-9aef-11ef-9cfb-f8b61b0dd1db

See the SIP traces - before the ACK was sent to 127.0.0.1, causing the call to fail.

### DIFF
--- a/sip/request.go
+++ b/sip/request.go
@@ -230,14 +230,21 @@ func NewAckRequest(inviteRequest *Request, inviteResponse *Response, body []byte
 	)
 	ackRequest.SipVersion = inviteRequest.SipVersion
 
+	destination := inviteRequest.Destination()
 	if len(inviteRequest.GetHeaders("Route")) > 0 {
 		CopyHeaders("Route", inviteRequest, ackRequest)
 	} else {
 		// https://datatracker.ietf.org/doc/html/rfc2543#section-6.29
-		hdrs := inviteResponse.GetHeaders("Record-Route")
-		for i := len(hdrs) - 1; i >= 0; i-- {
-			recordRoute := hdrs[i]
-			ackRequest.AppendHeader(NewHeader("Route", recordRoute.Value()))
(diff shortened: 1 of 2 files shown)


emiago:
Hi @DominikRidjic I think you found corner case. This was handled under TransactionRequest but seems it needs similar logic for ACK. 
I need to refactor this. This ack request building is also used for non 2xx

sip/request.go
```diff
@@ -230,14 +230,21 @@ func NewAckRequest(inviteRequest *Request, inviteResponse *Response, body []byte
...
 		// https://datatracker.ietf.org/doc/html/rfc2543#section-6.29
-		hdrs := inviteResponse.GetHeaders("Record-Route")
-		for i := len(hdrs) - 1; i >= 0; i-- {
-			recordRoute := hdrs[i]
-			ackRequest.AppendHeader(NewHeader("Route", recordRoute.Value()))
+		recordRoutes := inviteResponse.GetHeaders("Record-Route")
+		for i := len(recordRoutes) - 1; i >= 0; i-- {
+			ackRequest.AppendHeader(NewHeader("Route", recordRoutes[i].Value()))
+			// we need to set destination to last record-route
+			if i == len(recordRoutes)-1 {
```
emiago, line 242: I think here is more problem of forced destination

PR #152, opened 2024-11-11
response should include contact

wuxingzhong:


emiago:
In case of dialog, yes but not like this. Use dialog managment to respond or send requests

PR #155, opened 2024-11-26
Add support for SIP SRV record lookup

habakke:
Add support for looking up SIP SRV records when they exist to correctly route SIP traffic.

_sip._tcp.<host> for SIP
_sips._tcp.<host> for SIP

According to RFC3263


emiago:
hi @habakke . Thanks for looking into this. SRV lookup already exists and it is by default on. Please open issue to discuss your problem.

emiago:
Closing this as fixed, or until feedback is received otherwise.

PR #156, opened 2024-11-27
fix: add sip uri port if missing

InnovativeUS:
Add default port if it is missing in an address.

This is needed to ensure the requests without a port are matched in the connection pool


emiago:
Default port is handled by transport layer, so if you send request and port is missing transport layer will add it. 
This comes couple times already, but library has no usage of this method. It was mostly added as read property, but sip URI has limited knowledge of what transport will be used?

InnovativeUS:
Unfortunately if the request has no route, the address of the contact is used. 

sipgo/dialog_server.go:144

	// Check Route Header
	// Should be handled by transport layer but here we are making this explicit
	if rr := req.Route(); rr != nil {
		req.SetDestination(rr.Address.HostPort())
	}

That was where we were experiencing the issue

emiago:
Hi @InnovativeUS can you actually test your code, by removing that line. Normally Route headers are handled by default. 
Problem is, I can not get reason anymore why I made this explicit. Probably there was old issue

emiago:
Hi @InnovativeUS . Not sure do I get your last response correctly. If anything needs to done here, go ahead open issue. 
I will check that line my self once more.

Issue #157, opened 2024-11-28
Any plan to add Refer Header?


kota-yata:
I see there is method constant REFER, but header structs for neither refer-to nor referred-by are provided. is there any plan to add those?

kota-yata:
I would love to contribute to add those struct as transferring phone calls is kind of a key component of my application

emiago:
@kota-yata go ahead, if it is just adding structs.

PR #158, opened 2024-11-30
Add Header structs and parsers for Refer-to and Referred-by

kota-yata:
from #157 

Refer-to: [rfc3515](https://datatracker.ietf.org/doc/html/rfc3515)
Referred-by: [rfc3892](https://datatracker.ietf.org/doc/html/rfc3892)

for Refer-to parser, it calls `parseRouteAddress` instead of creating new one because the structure is same.
for Referred-by parser I simply copied ToHeader and FromHeader's parsing logic.

### DIFF
--- a/sip/headers.go
+++ b/sip/headers.go
@@ -49,6 +49,7 @@ type headers struct {
 	route         *RouteHeader
 	recordRoute   *RecordRouteHeader
 	maxForwards   *MaxForwardsHeader
+	referTo       *ReferToHeader
 }
 
 func (hs *headers) String() string {
@@ -94,6 +95,8 @@ func (hs *headers) setHeaderRef(header Header) {
 		hs.contentType = m
 	case *MaxForwardsHeader:
 		hs.maxForwards = m
+	case *ReferToHeader:
+		hs.referTo = m
 	}
 }
 
@@ -121,6 +124,8 @@ func (hs *headers) unref(header Header) {
 		hs.contentType = nil
 	case *MaxForwardsHeader:
 		hs.maxForwards = nil
+	case *ReferToHeader:
+		hs.referTo = nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kota-yata:
@emiago 
ToHeader, FromHeader and ContactHeader have the same header structure and only ContactHeader has different parse logic than other two: https://github.com/emiago/sipgo/blob/4fd9a96f049e4066a6ff8aa169cdc75aec92b449/sip/parse_address.go#L220

is there any reason for this?

sip/headers.go
```diff
@@ -121,6 +127,10 @@ func (hs *headers) unref(header Header) {
 		hs.contentType = nil
 	case *MaxForwardsHeader:
 		hs.maxForwards = nil
+	case *ReferToHeader:
```
emiago, line 130: Fast references are only for headers that you need to have or read. 
Please remove this

sip/headers.go
```diff
@@ -407,6 +417,28 @@ func (hs *headers) RecordRoute() *RecordRouteHeader {
 	return hs.recordRoute
 }
 
+// ReferTo returns underlying Refer-To parsed header or nil if not exists
+func (hs *headers) ReferTo() *ReferToHeader {
```
emiago, line 421: you can keep lazy parsing, but you need to remove this fast references.

emiago (changes requested):

sip/headers.go
kota-yata: do you mean automatic assignment to h.referTo/h.referredBy need to be removed? I don't get what you mean by "fast references"

sip/headers.go
kota-yata: removed statement from both setHeaderRef and unref

sip/headers.go
emiago: yes.

emiago:
> @emiago ToHeader, FromHeader and ContactHeader have the same header structure and only ContactHeader has different parse logic than other two:
> 
> https://github.com/emiago/sipgo/blob/4fd9a96f049e4066a6ff8aa169cdc75aec92b449/sip/parse_address.go#L220
> 
> is there any reason for this?

Everything has the same parse func, not sure what are you targeting. 
Please be consistent with other API. JUst checked you are returning bool for some reason. nil check is enough

kota-yata:
might be confused with Rust syntax at the time. fixed

emiago:
Hi @kota-yata I merged this. I hope you do not mind I rebased and squashed your commits.

PR #161, opened 2024-12-05
add storage for user business data in dialog

wuxingzhong:
Enhance user data storage capability in dialog, allowing some data during invite to be stored in the dialog and used during ack and bye.

Or see if there is a more reasonable solution.

### DIFF
--- a/dialog.go
+++ b/dialog.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 	"sync/atomic"
 
 	"github.com/emiago/sipgo/sip"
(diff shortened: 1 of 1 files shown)


emiago:
Hi @wuxingzhong , thanks for mentioning this.
Why just not wrap dialog with this. Dialogs are composable and all calls are explicit. Please open issue and describe why this can not be solved in other way

emiago:
Hi @wuxingzhong , those are just library default caches. They are not necessary. You can build your own caches, with small effort. 
Maybe check then diago project, which can have cache replaced (generic implementation). To pass this custom values, you would need just to extend your cache functionality, and have updates.

PR #163, opened 2024-12-12
add user-agent header in register sip msg.

daimon99:
With this patch, you now can add User-Agent header in register sip message.
	
ua, _ := sipgo.NewUA(
		sipgo.WithUserAgent(username),
		sipgo.WithUserAgentHeader(useragent),
		sipgo.WithUserAgentHostname("localhost"),
	)

### DIFF
--- a/client.go
+++ b/client.go
@@ -256,9 +256,10 @@ func clientRequestBuildReq(c *Client, req *sip.Request) error {
 	// https://www.rfc-editor.org/rfc/rfc3261#section-8.1.1
 	// A valid SIP request formulated by a UAC MUST, at a minimum, contain
 	// the following header fields: To, From, CSeq, Call-ID, Max-Forwards,
-	// and Via;
+	// Via, and User-Agent;
 
-	mustHeader := make([]sip.Header, 0, 6)
+	mustHeader := make([]sip.Header, 0, 7)
+
 	if v := req.Via(); v == nil {
 		// Multi VIA value must be manually added
 		via := clientRequestCreateVia(c, req)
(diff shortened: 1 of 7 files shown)


emiago:
Hi @daimon99 . Thanks for taking you time and effort on making this. 

Still I see here breaking change `WithUserAgent(username, uaHeader string) UserAgentOption` 

On other hand this is not per RFC required header, although sometimes I see mostly provided. 
I would like to avoid polluting SIP lib with optional things, but I understand this. 

I need to think about this is it worth merging. Still consider to fix that break change, as it does not match your comment?
There is also no need to create special header type, or do you see different? There is GenericHeader type for all this string.

Fast reference is also thing we need to avoid adding (memory allocs -> GC)

daimon99:
Sorry, I accidentally left out a part of the code in my previous submission. Additionally, I did not notice the existence of the NewHeader method. The background for this PR is as follows:

    The current withUserAgent(name) method can be misleading, as it suggests that it constructs a header with the User-Agent information, but in fact, it does not. This can cause confusion. I only realized after reading the source code that this refers to the user's authentication username, not the value of the User-Agent.

    I need to submit a User-Agent header in the register method, but I couldn't find a corresponding method in the existing framework, so I ended up reinventing the wheel. Following your advice, I'll check again to see if the current framework can meet my needs. If necessary, I will submit another PR.

I agree with your performance-first approach; it should always have a purpose.

emiago:
Hi @daimon99 , 

Just to be clear this is your user to identify on From. Could be called just User, but it is not Username for auth. 
On digest you can have different username. This is just your some ID. 
So I agree naming could be better, probably as UserID, UserHostname. 

I may see like need to provide this kind of additional headers on each request, so I would like to think this feature more from client perspective. In short anyone can wrap client `Do, TransactionRequest` but could be worth a while to have it in a lib.

emiago:
we will handle this as issue. Most probably we will make this simple and avoid creating special header entry.

PR #170, opened 2024-12-30
feat: add NewXxx for the type-aliased headers

hateeyan:
Hi,

This PR add NewXxx for the type-aliased headers so that we don't need to create variables separately when appending headers.


hateeyan:
Hi @emiago , thanks for your reply.

I understand the problem now. There may be no good solution at present. I am currently using `sip.NewHeader` instead of initializing variables. We may find a better way to solve it, or keep it as it is.

PR #174, opened 2025-01-08
Fix TODO for non-routable contact address in BYE

juberti-oai:
cc @juberti
When debugging 1-800-CHATGPT, I noticed that we would fail to send SIP BYE if the contact header was non-routable (for some providers we get a 10.x.x.x address). There was an existing TODO to handle this, so I implemented the necessary logic.

I also made a nearby change to set the BYE transport to match what is specified in the Contact header. 

All existing tests pass.

### DIFF
--- a/dialog_server.go
+++ b/dialog_server.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


emiago:
Hi @juberti-oai . Thanks for contributing. 
This is probably breaking RFC. 
Check option for RewriteContact. This forces to use source as destination. 
This is added to fix NAT issue. It should be familiar like in other VOIP tech.

For this comment in code I myself have to check. Maybe it was some left over. :) 
Regards to ChatGPT team!

juberti-oai:
Thanks for the quick reply. Let me try that option, that looks like what I was looking for.

juberti-oai:
Thanks for pointing out this flag, it solved the issue we were hitting. Closing this PR accordingly.

I'm curious, is rewriting the contact like this RFC-compliant?

PR #180, opened 2025-02-03
Put Route headers in the correct order

juberti-oai:
We encountered an issue when onboarding a new telephony provider, where sipgo would send BYE to the wrong proxy. Debugging revealed an INVITE of the form

```
INVITE sip:REDACTED@REDACTED:5060 SIP/2.0
Record-Route: <sip:192.76.120.10;transport=tcp;r2=on;lr;ftag=t6Z39cN3Hg3DQ>
Record-Route: <sip:10.255.0.1;r2=on;lr;ftag=t6Z39cN3Hg3DQ>
Via: SIP/2.0/TCP 192.76.120.10;branch=z9hG4bKbab1.886142c07f4574308096558d6593615a.0
v:SIP/2.0/UDP 10.13.211.24:6000;received=10.13.211.24;rport=6000;branch=z9hG4bKD53H2Zetjg6vF
<snip>
```

but the BYE was being sent to 10.255.0.1 instead of 192.76.120.10, suggesting an issue with the processing of the `Record-Route` headers. 

I looked into the sipgo code to understand why this occurred and for some reason the code is reversing the order of the headers when writing out the `Route` headers. As indicated in [RFC 3261, S 16.12.1.2](https://datatracker.ietf.org/doc/html/rfc3261#section-16.12.1.2) (the section mentioned in the code comment), the order of the `Route` headers should match the order of the `Record-Route` headers.

This 1-line change updates `dialog_server.go` to correct the `Route` header ordering.

### DIFF
--- a/dialog_client.go
+++ b/dialog_client.go
@@ -153,7 +153,7 @@ func (s *DialogClientSession) buildReq(req *sip.Request) {
 		// More on
 		// https://datatracker.ietf.org/doc/html/rfc3261#section-16.12.1.1
 		if len(hdrs) > 0 {
-			for i := len(hdrs) - 1; i >= 0; i-- {
+			for i := range hdrs {
 				// We need to put record-route as recipient in case of strict routing
 				recordRoute := hdrs[i]
 				req.AppendHeader(sip.NewHeader("Route", recordRoute.Value()))
--- a/dialog_client.go
+++ b/dialog_client.go
@@ -153,7 +153,7 @@ func (s *DialogClientSession) buildReq(req *sip.Request) {
 		// More on
 		// https://datatracker.ietf.org/doc/html/rfc3261#section-16.12.1.1
 		if len(hdrs) > 0 {
-			for i := range hdrs {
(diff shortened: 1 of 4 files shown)


emiago:
nice. but I would like finally to wrap this with test. Will open issue. Dialog was done with some refactoring and I wonder is this due to copy paste. My use case is single RR so could be reason why never catched.

juberti-oai:
I fixed the tests in `dialog_server_test.go` and also in `dialog_client_test.go` (and fixed the same bug in `dialog_client.go`). Let me know if there's anything else I can help with.

emiago:
Hi @juberti-oai . No please revert. Dialog client is a mistake, it should do reverse. Leave last 2 commits out, I can recheck tests. I want to also check did we do something wrong more.

juberti-oai:
OK, reverted. Let me know if you need anything else from me.

emiago:
merged

juberti-oai:
Thanks, please let me know when you cut a release so we can pick it up.

Issue #57, opened 2023-11-09
rfc4475 test


emiago:

darkrift:
Torture test setup with tests from RFC4475 was brought in with https://github.com/emiago/sipgo/pull/192

PR #192, opened 2025-02-17
bring torture test setup

darkrift:
This brings the torture test messages from RFC4475 and bring part of the solution for #57

They are grouped according to their state (valid/invalid) from the RFC.

The currently failing tests are commented out as not to break go tests and releases.

They are also added to gitattribute so they keep their proper CR and CRLF ending as defined by the RFC using their method of extraction.

This can then become the base upon which to improve the library and make a better parser.


emiago:
hi @darkrift . Seems like great addition. I appreciate help. I only wonder could be better way to put all this tests and avoid many files, but I also see it may simple todo. 
Anyway I would check it out once have more time.

darkrift:
@emiago To be honest, I think we should keep it as it is. 

If you look at the data provided at the end of the RFC4475, the base64 data once decrypted results in a tar.gz with all those files in it as is, each describing a different scenario.

The only "processing" I did was to group them according to their valid state (section 3.1.1) or their invalid state (section 3.1.2)

That will also help tackling each test individually so it's easier to get fix.

darkrift:
@emiago Any chance we can get this merged or are you still looking at some alternative way to provide those tests ? 

Merging this would help people see what's the current parser standing with regards to the torture tests and let them jump in in order to help tackle the commented/failing tests.

We could probably open an issue for each of the failing tests after this is merged and close #57

emiago:
Hi @darkrift . Yeah PR on first sight should be merged, just had no time to look at it. 
Lets not open every issue. It is enough to keep one open, just document list on original issue, and we track changes.

sip/torture_test.go
```diff
@@ -0,0 +1,92 @@
...
+		"test",
+		"trws",
+		//"unkscm",
+		//"unksm2",
+		//"zeromf",
+	}
+	for _, test := range validTests {
+		t.Run(test, func(t *testing.T) {
+			data, err := os.ReadFile("testdata/torture/valid/" + test + ".dat")
+			assert.NoError(t, err)
```
emiago, line 72: please use here `require.NoError`

sip/torture_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			_, err = parser.ParseSIP(data)
+			if err != nil {
```
emiago, line 75: why mixing If statements and doing `require checks.` 
Just use `require.NoError` For more info you can pass at end insteaed using `fmt.Sprintf`

sip/torture_test.go
```diff
@@ -0,0 +1,92 @@
...
+				require.NoErrorf(t, err, fmt.Sprintf("error parsing %s", test))
+			}
+		})
+	}
+
+	for _, test := range invalidTests {
+		t.Run(test, func(t *testing.T) {
+			data, err := os.ReadFile("testdata/torture/invalid/" + test + ".dat")
+			assert.NoError(t, err)
```
emiago, line 84: use require.NoError to stop further execution.

sip/torture_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			_, err = parser.ParseSIP(data)
+			if err == nil {
```
emiago, line 87: again no need todo If and then require. Just use `require.NoError`. Pass additional message/info as last argument

emiago:
Hi @darkrift . I did review, can you maybe fix this error checking in test and then this should be good to merge.

darkrift:
@emiago Should be fixed

emiago:
Hi @darkrift . Thanks. This now gives great benefit to lib. I will look all this tests

PR #193, opened 2025-02-18
Update server.go, bug with writing to a channel, but it's function

Shegl:
found a bug trying to listen TLS on latest version of diago, sorry for not doing semver


emiago:
nice

PR #197, opened 2025-02-20
fix: ignore extra INVITE request prior 100 Trying

rianby64:
Based on [RFC 3261 17 Transactions](https://datatracker.ietf.org/doc/html/rfc3261#section-17), a retransmission of an INVITE should be ignored, as per:

```
   Similarly, the purpose of the server transaction is to receive
   requests from the transport layer and deliver them to the TU.  The
   server transaction filters any request retransmissions from the
   network.
```

### DIFF
--- a/sip/transaction_server_tx.go
+++ b/sip/transaction_server_tx.go
@@ -59,7 +59,7 @@ func (tx *ServerTx) Init() error {
 		tx.timer_1xx = time.AfterFunc(Timer_1xx, func() {
 			trying := NewResponseFromRequest(
 				tx.Origin(),
-				100,
+				StatusTrying,
 				"Trying",
 				nil,
 			)
@@ -79,7 +79,7 @@ func (tx *ServerTx) Init() error {
 // therefore running in seperate goroutine is needed
 func (tx *ServerTx) Receive(req *Request) error {
 	tx.mu.Lock()
-	if tx.timer_1xx != nil {
+	if tx.timer_1xx != nil && !req.IsInvite() {
 		tx.timer_1xx.Stop()
 		tx.timer_1xx = nil
 	}
--- a/sip/transaction_server_tx_test.go
+++ b/sip/transaction_server_tx_test.go
@@ -28,11 +28,14 @@ func TestServerTransactionFSM(t *testing.T) {
 		tx := NewServerTx("123", req, conn, slog.Default())
 		err := tx.Init()
 		require.NoError(t, err)
+		require.NoError(t, tx.Err())
 
 		err = tx.Receive(req)
 		require.NoError(t, err)
-
 		require.NoError(t, tx.Err())
+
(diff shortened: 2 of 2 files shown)


rianby64:
Hello @emiago 

I was playing a bit with baresip and other things, and I decided to check one case... what IF I send several times the same INVITE? What I found was a `transaction terminated`. I think a retransmission (sending the same INVITE by error or by reliability) should not break the thing. What do you think?

emiago:
Hi @rianby64 . Library (transaction layer) handles re transmission. That is why you have this `Done()` channel to be signaled if terminated. It is based on SIP timers.
I do not understand why you as caller should send same INVITE, but that is not that you as caller should DO.

emiago:
Hi @rianby64 this must be then some bug. I will check

emiago:
Hi @rianby64 thanks for catching this. I have fixed now. Wow

emiago:
Can you see does this solve issue on your end. If I saw, we did return error when there was no response set, and we want to pass one. 
This could happen on initial request.

rianby64:
Yeah! thanks a lot! @emiago nice and elegant solution!

Now... let's check one last case.
If a second(same) INVITE is sent to the server, I think the best thing is to ignore it. So, I noticed `100 Trying` is being stopped because of the second undesired INVITE. In this case (with the second INVITE) we are affecting the state of the machine.

```go
	if tx.timer_1xx != nil {
		tx.timer_1xx.Stop()
		tx.timer_1xx = nil
	}
```

I updated my pull request with one last thing.
I was thinking about `tx.timer_1xx.Reset(Timer_1xx)` but looks to me like a place that can exploit the state of `100 Trying`, like sending the same INVITE 2500 times in a row will block the timer. Anyway.

This is the result from my updated pull request:
image

And here is from your solution:
image

What about setting the timer from the FIRST INVITE to be triggered by the expected 200 ms, as per the standard?
I really appreciate your opinion regarding this small final detail.

rianby64:
However... https://datatracker.ietf.org/doc/rfc3665/ _section 3.1 Successful Session Establishment_ shows a different story.

Seems to me like sending a second INVITE in a raw breaks the authorization. What about if the INVITE requires authorization?

Now I'm a bit confused with this, but still I stand firmly on the idea not to stop the timer because of the second INVITE.

Anyway :slightly_smiling_face: wish you the best with this question.

emiago:
Hi @rianby64 . You are putting great points, but lets not forget, that in then end you are controling transaction fully, and you can send provisional responses. So if I remember if you pass your prov response, after that each retransmission will get immediatelly response, which actually was causing issue in first place by missing prov response.
This timer is only there in case you are just too slow, so I am not sure should this be more complicated.
Please try to see what happens when you pass prov response imediately to your transaction and how this will be handled.

rianby64:
Ah yes! Agree :+1: 
Now I can close this ticket.

New cases may arise, then new issues we can open.

Thank you for this conversation and solution.

Have a nice day!

Issue #206, opened 2025-03-12
Handling Large SIP INVITE Packets in UDP


pavanputhra:
#### Description  
I am trying to send an INVITE request whose size is around 4000 bytes. However, I receive the following error:  

```
{"level":"debug","caller":"transactionLayer","error":"size of packet larger than MTU","req":"INVITE sip:foo@example.com SIP/2.0","time":"2025-03-12T04:47:50Z","message":"Fail to write request on init"}

size of packet larger than MTU. transaction transport error
```  

I am using the following code to forward the SIP request over UDP:  

```go
clTx, err := s.client.TransactionRequest(ctx, req, sipgo.ClientRequestAddVia, sipgo.ClientRequestAddRecordRoute)
```  

#### Question  
Is there a way in the `sipgo` library handle large SIP messages by automatically fragmenting them when using UDP? If not could this feature be implemented?

#### Note
I am aware that the SIP RFC recommends switching to TCP when a message exceeds the MTU. However, some proxies like Kamailio implement UDP fragmentation in such situations, which is useful when there is no reliable way to determine whether the User Agent supports TCP.

Thanks!

emiago:
Hi @pavanputhra . To keep it short, I think this will not happen, but as library is flexible, there could be a way that you could override default transport implementation. Of course if you are willing to go this path, we could consider exposing this. 

BTW: It is not recommends. -> https://datatracker.ietf.org/doc/html/rfc3261#section-18.1.1
Kamailio can do automatic switch as well.

PR #208, opened 2025-03-15
Add support for UDP fragmentation for large SIP messages

pavanputhra:
This PR addresses issue #206 by enabling UDP fragmentation for SIP messages that exceed the MTU size.

### DIFF
--- a/sip/sip.go
+++ b/sip/sip.go
@@ -13,6 +13,7 @@ const (
 
 var (
 	SIPDebug  bool
+	AllowUDPFragmentation bool = false
 	siptracer SIPTracer
 )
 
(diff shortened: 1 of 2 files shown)


emiago:
Hi @pavanputhra . This does not solve any issue? You have UDPMTUSize as well. Please explain

pavanputhra:
@emiago

The goal was to maintain the existing behavior, which is why UDPMTUSize remains in place—ensuring that library users can adhere strictly to the RFC. At the same time, they have the flexibility to bypass this limitation by setting AllowUDPFragmentation to true. 

Do you have any other implementation suggestions?

emiago:
Hi @pavanputhra . I see no reason why would someone use both values. This is reason why value exists. You can set it bigger if you want. Maybe just try and if there is specific issue please open.

pavanputhra:
I get it. It we could set `UDPMTUSize=99999` get the same desired effect. Closing the PR.

PR #209, opened 2025-03-16
Add user agent logging option

InnovativeUS:
It is possible to replace the logging in the library using WithServerLogger and WithClientLogger. This pull request adds an equivalent WithUserAgentLogger to replace the logger for the transaction and transport layer that are created by the UA. 

In addition, it ensures that the transports use the logger when debugging trace messages so that they are output to the transport and transaction logger rather than the global slog.

### DIFF
--- a/sip/transaction_layer.go
+++ b/sip/transaction_layer.go
@@ -29,11 +29,13 @@ type TransactionLayer struct {
 	log *slog.Logger
 }
 
-type TransactionLayerOption func(tpl *TransportLayer)
+type TransactionLayerOption func(tpl *TransactionLayer)
 
 func WithTransactionLayerLogger(l *slog.Logger) TransactionLayerOption {
-	return func(tpl *TransportLayer) {
-		tpl.log = l
+	return func(txl *TransactionLayer) {
+		if l != nil {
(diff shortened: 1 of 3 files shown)


emiago:
Hi @InnovativeUS . We do not need logger pointer everywhere unless it is really needed. I would like to be careful with this due to passing this reference, and rather consider alternative.
Instead as you should notice there is SIPTracer which you can override and do own logging. 

I am not against exposing way to customize logger to transaction layer. This can be exposed.

InnovativeUS:
Ok

Thank you for the guidance. We will update the PR. Are you ok with adding an option for the tracer object so that we avoid globals?
________________________________

InnovativeUS:
We have removed the pointers in the transport logger and separated out Transport and Transaction logger so that they can be set when creating a UserAgent. Let us know if you need any other changes

emiago:
please squash your changes.
Actually we need todo this differently. We should allow passing transport and transaction options. This way you could manipulate this from caller side and it will automatically solve problem of new options.

emiago:
hi @InnovativeUS . Yeah, it just feels this will give better control to caller.

InnovativeUS:
Hopefully this will work for you

Issue #207, opened 2025-03-13
Issue: Cannot Bind Public IP in EC2 When Adding Via Header


pavanputhra:
#### Description  
I am encountering an issue when deploying a SIP proxy build using the `sipgo` library on an EC2 instance.  

When sending a SIP message out to the internet, I want to include a `Via` header that contains the public IP of the instance. However, this results in the following error:  

```
listen udp publicIp:5060: bind: cannot assign requested address
```

The reason for this error is that the public IP assigned by AWS is not directly bindable. Instead, I must bind to `0.0.0.0` or the private IP assigned to the instance.  

Upon inspecting the `sipgo` code, I noticed that the `ClientRequestConnection` function relies on the `Via` header to determine the local address binding. This behavior causes the issue in an EC2 environment where the public IP cannot be used for binding. (This is also true while running app in containerised environment)

#### Relevant Code Section  
```go
// Now use Via header to determine our local address
viaHop := req.Via()
if viaHop == nil {
    return nil, fmt.Errorf("missing Via Header")
}

laddr := Addr{
    IP: net.ParseIP(viaHop.Host),
    Port: viaHop.Port,
}
```

We need some other way to determine `laddr` without tightly coupling to `Via` header.

emiago:
yes, this must be found better way. as it is recommended per RFC FQDN could be better approach, and unless something did not broke, this will actually allow you to bypass this issue.

pavanputhra:
@emiago There are two issue with using FQDN in my case

1. During development, in my local computer I am using dockerised environment. Using an FQDN in the Via header is not possible or complicated.

2. I explored the sipgo code to find which IP it binds to when an FQDN is used in Via, but I couldn’t find it. If the code resolves the FQDN to a public IP and tries to bind to it, then the issue remains the same.

pavanputhra:
@emiago I attempted to use an FQDN to work around this restriction, but unfortunately, I encountered the following error:  

```
Failed to create client transaction: listen udp4 :5060: bind: address already in use
```  

### Proposed Solution  
After analyzing the issue, I came up with a potential solution. Let me know your thoughts on this approach.  

We could introduce an additional field in the `UA` struct, named `viaIpToBindIpMap`. This would act as a mapping from `publicIp` to `privateIp`, determining the appropriate IP to bind to for a given `Via` header IP.  

#### Example Implementation  

In the `UA` struct, we could define:  

```go
type UA struct {
[...]

PR #214, opened 2025-03-21
Issue 207: Add Request.AdvertisedHost Field

pavanputhra:
This PR implements Request.AdvertisedHost, as suggested in [this comment](https://github.com/emiago/sipgo/issues/207#issuecomment-2733100760) for issue #207.

The concept of AdvertisedHost is inspired by how Kamailio handles it in its configuration:

```ini
listen=udp:0.0.0.0:5060 advertise PUBLIC_IP:5060 name "udp_public"  
```

@emiago Let me know if this approach aligns with your expectations.

### DIFF
--- a/sip/message.go
+++ b/sip/message.go
@@ -131,9 +131,9 @@ type MessageData struct {
 	tp         string
 
 	// This is for internal routing
-	src    string
-	dest   string
-	ViaNAT bool
+	src            string
+	dest           string
+	AdvertisedHost string
 }
 
 func (msg *MessageData) Body() []byte {
--- a/sip/transport_layer.go
+++ b/sip/transport_layer.go
@@ -417,6 +417,9 @@ func (l *TransportLayer) ClientRequestConnection(ctx context.Context, req *Reque
 				viaHop.Host = host
 			}
 			viaHop.Port = port
+			if req.AdvertisedHost != "" {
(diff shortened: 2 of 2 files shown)


emiago:
hmm. Ok I now see your idea. It could be better way yes. Just I am trying to think can we find better naming. 
As I see this more related to transport. I would like to also challenge now this concept, I have
When Via is used:
- Retransmission or sending same request is fine from lib, but could be issue from client (binding issue) 
- Binding mostly has sense for UDP, but not for other protocols. Connections are attached to Raddr. Again using here Via could be issue. 

Alternative:
Forcing Binding via special `request.BindHost`
- We check does connection exist only int this case. Mostly UDP
- We do not use Via sent by to check connection as mostly it will not be case. We cache connection based on `raddr` with UDP exception. 
- Should we still force laddr when present in Via? In this case sent-by is not honored, but as this mostly lands to be FQDN or Public IP it could make more sense. 
- Client Host and port is replaced with BindHost, and Advertising is actually just done on Via header. This will be big break change, but we can keep backward compatibility and extend with Advertise

@pavanputhra. How about this. Thanks for challenging?

emiago:
also we should use here Setter, as it could be better to store IP + Port instead string. Benchmarking should be done

emiago:
fixed as explained in #207

Issue #218, opened 2025-03-31
Add a method to get the local address port (Expose Server Transaction Connection)


hateeyan:
Hi, 

When listening on multiple addresses and ports using the same transport, it is currently impossible to know which local address and port the request came from.

```go

	server, err := sipgo.NewServer(ua)
	if err != nil {
		panic(err)
	}

	go func() {
		if err := server.ListenAndServe(context.TODO(), "udp", "192.168.1.1:5060"); err != nil {
			panic(err)
		}
	}()
	go func() {
		if err := server.ListenAndServe(context.TODO(), "udp", "192.168.2.1:5070"); err != nil {
			panic(err)
		}
	}()
```
Maybe add a method to get the local address port in `MessageData`?

hateeyan:
Hi @emiago ，

I am working on a SIP proxy with dual NIC. After receiving a SIP message on one NIC, it needs to be sent from another NIC.
So I need to know the local address of the SIP message to know which NIC it was received from.

emiago:
If we expose `Connection` on `ServerTx` would for you be ok, that as temporally solution, you get this by reading from Connection property. 
I think you have special case, and I just want to be careful on adding more properties on request.
Normally routing in proxy takes path of remote address checking, or some way authorizing request.

hateeyan:
> If we expose `Connection` on `ServerTx` would for you be ok, that as temporally solution, you get this by reading from Connection property.

Thanks for reply @emiago , that's what I needed!



> I think you have special case, and I just want to be careful on adding more properties on request. Normally routing in proxy takes path of remote address checking, or some way authorizing request.

Yes, it is a special case, dual NIC with NAT traversal.

emiago:
done.

PR #219, opened 2025-04-09
feat: expose LocalAddr for Transaction

hateeyan:
Hi, @emiago 

According to the discussion in the issue, this PR only exposes the `LocalAddr` method, not the complete `Connection`.

Close: https://github.com/emiago/sipgo/issues/218

### DIFF
--- a/sip/transaction.go
+++ b/sip/transaction.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"reflect"
 	"runtime"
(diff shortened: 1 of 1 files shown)


emiago:
yes, but it is now bad abstraction. I do not want to have more here things, or at least it is better return larger objects. Lets just make Connection method, and you can fetch this with type conversion. It is not something unseen in go world. You have this a lot with HTTP casting.
If this really makes sense to expose, I would do this, but we need more usage examples.

PR #232, opened 2025-05-22
feat: enable to customize TCP dialer

higebu:
Example:

```go
control := func(network, address string, c syscall.RawConn) (err error) {
	controlErr := c.Control(func(fd uintptr) {
		err = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEADDR, 1)
		if err != nil {
			return
		}
		err = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT, 1)
	})
	if controlErr != nil {
		err = controlErr
	}
	return
}
dialer := &net.Dialer{
	Control: control,
}
ua, _ := sipgo.NewUA(
	sipgo.WithUserAgentTransportLayerOptions(sip.WithTransportLayerTCPDialer(dialer)),
)
```

### DIFF
--- a/sip/transport_layer.go
+++ b/sip/transport_layer.go
@@ -26,6 +26,8 @@ type TransportLayer struct {
 	ws  *transportWS
 	wss *transportWSS
 
+	tcpDialer *net.Dialer
+
 	transports map[string]Transport
 
 	listenPorts   map[string][]int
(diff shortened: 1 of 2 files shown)


emiago:
Please open issue. I would see to put it on higher priority.

Issue #242, opened 2025-06-22
Allow creation of DialogClientSession without sending an INVITE


dgleiche:
I can see no way to create a `DialogClientSession` outside of calling `WriteInvite` on a `DialogUA`, since certain necessary fields on `DialogClientSession` are private (such as `ua`). We sometimes need to create a `DialogClientSession` after the dialog has already been established in order to send out requests after failing an active call over to a different UA server, for example. 

Happy to submit a PR on this, but wanted to check and see if this sounds reasonable, or if there's something I missed which allows a `DialogClientSession` to be created without sending an INVITE out. We're able to get around this on `DialogServerSession` and recover in-progress calls by using `ReadInvite` with the initial request that was received (which we save to a DB) and providing a no-op server transaction, but this is still a bit awkward.

emiago:
Hi @dgleiche . It is doable. Please check diago library and function called NewDialog. I am pretty sure you are looking old code.

dgleiche:
Thank you, I missed that UA is now public. Two more related questions:
- diago cannot set `inviteTx` on the underlying `sipgo.Dialog`. Won't this cause nil-pointer exceptions? (e.g. here: https://github.com/emiago/sipgo/blob/c76d432a3e984c548a0105aa11d306cf9b8eecc8/dialog_client.go#L431)
- diago does not allow us to set the CSeq number when recovering a dialog. We get around this now by setting the latest CSeq number on the aforementioned `ReadInvite` call, but this is hacky. 

Any chance we can make the above fields public too?

emiago:
Hi @dgleiche. 
I do not understand your use case, diago session is not active session until gets dialed. 
What do you mean by recovering dialog, I hope you are not doing this on transactional level?

dgleiche:
> Hi @dgleiche. 
> I do not understand your use case, diago session is not active session until gets dialed. 
> What do you mean by recovering dialog, I hope you are not doing this on transactional level?

I'm referring to recovering a dialog after the initial INVITE, e.g. one server handles the initial INVITE transaction and another server handles the transaction for sending the BYE.

dgleiche:
@emiago I'd be happy to post a PR on this if you'd be open to that. I'm anticipating the necessary changes should be pretty non-controversial (just adding some nil checks and a function to set the lastCSeq).

PR #244, opened 2025-06-24
Add ad-hoc dialog creation

dgleiche:
Related to: https://github.com/emiago/sipgo/issues/242.

Allows the creation of `DialogClientSession` / `DialogServerSession` after the initial transaction has been completed. This is useful in failover and redundancy scenarios where different transactions within a dialog may be handled by different servers.

### DIFF
--- a/dialog.go
+++ b/dialog.go
@@ -144,3 +144,7 @@ func (d *Dialog) CSEQ() uint32 {
 func (d *Dialog) Context() context.Context {
 	return d.ctx
 }
+
+func (d *Dialog) SetCSEQ(cseq uint32) {
+	d.lastCSeqNo.Store(cseq)
+}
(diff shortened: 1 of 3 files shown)


emiago:
Hi @dgleiche . Now please rethink about this, and consider what actually you really need. 
Side NOTE: It is fine that you develop side things with lib. I do not push all stuff to lib as well ;)

emiago:
Hi @dgleiche . Will be SetCSEQ be enough for your case. Sorry but I must say I will not support constructors in this way. 
There are `Init` functions and this will be mostly used method for now. 
I see potential making CSEQ customizable. 
I would like to understand how and why are you reconstructing transactions, as this is overkill for dialogs except if this is only on init.
Nil pointer risk I understand, but I find hard that you can avoid calling ReadInvite. 

Lets move discussion again to issue, but lets do small steps. Thanks for sharing PR.

PR #246, opened 2025-07-09
feat: add compact headers transport option

InnovativeUS:


emiago:
Hi @InnovativeUS . Thanks for pushing this out.  This is unnecessary breakage I must say, that is I would only solve this on higher level. There is no need that transport layer has knowledge about this. 
I have solution in mind, but I can not guarantee that I will find any time soon. 
Your problem is obviously client side, as parsing could have workarounds. I know right now is not this exposed for better control. 
Anyway we need to solve both. 
It is some effort definitely, and also lets be aware compact headers are limited set. Benefits nowdays are low due to more secure transport. 

I will add this on roadmap before 1.0.

PR #253, opened 2025-08-27
fix: add content length header to bye

InnovativeUS:
When using TCP / TLS transport, when sending a BYE a warning is received that the content length header is not provided.

This pull request adds a Content-Length header of 0 to outgoing BYE requests.


emiago:
Thanks for PR, but to avoid back and forth, I added fix on transaction level. It was copy past from dialog client. 
 
PR is referenced. Closing  as fixed

PR #254, opened 2025-09-01
fix: ClientTx change to proceeding state can't stop timer_b.

zozobreakzou:
timer_b is timeout of the whole tranction including final response.

Bug Scenario: UAS response 1xx but no 2xx, the tx will still forever.


zozobreakzou:
image
image

PR #255, opened 2025-09-22
fix: ParserStream handling StartLine and incomplete Header

j-licht:
Before the fix, when ParserStream got an partial message only containing a StartLine and parts of the following header, it failed with 'field name with no value in header: EOF on reading line' We need to adjust the unparsed slice after reading the StartLine.


emiago:
Hi @j-licht , thanks for finding.
Can Test be with string buffers, I hardly see need to maintain like this for SIP cases. It would be easier to understand as well.

PR #263, opened 2025-11-06
Cancel req issues

disacod:
1. When processing Cancel, the first response is 487, then OK, which is incorrect.
2. When processing Cancel, the session hangs until the timeout with "WARN ACK missed." I think stateProceeding would be a better handler for fsm onCancel than inviteStateProcceeding.


emiago:
Hi @disacod .

Ok I see your point. Probably even better error handling should be done as well, but even if we fix order, your client should threat them as seperate transcations. This can indicate also bad client. 
OK is for CANCEL, 487 is for INVITE. 

I have to look more further, as it would be also wrong to say CANCEL is processed successfully?

emiago:
I think this majorly fixed now. In any case should be rebased @disacod and reopen

PR #268, opened 2025-12-01
transport_tcp: add read and write deadline

j-licht:
add read and write deadline for tcp transport, otherwise it takes a long time to detect offline devices.

### DIFF
--- a/fakes/tcp_conn.go
+++ b/fakes/tcp_conn.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"sync"
 	"testing"
+	"time"
 )
 
 type TCPConn struct {
@@ -48,6 +49,14 @@ func (c *TCPConn) Write(p []byte) (n int, err error) {
 }
 
 func (c *TCPConn) Close() error {
+	return nil
+}
+
+func (c *TCPConn) SetReadDeadline(time.Time) error {
+	return nil
+}
+
+func (c *TCPConn) SetWriteDeadline(time.Time) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


emiago:
I guess main issue could be now for client usage as you need to have this connection upfront. 

I find this very hardcoded. Could maybe TCP keep alive solve problem?

emiago:
Hi @j-licht . Right now you can customize transports trhough `TransportsConfig` which can be passed through `UA` on TransportLayer. 
What we basically need is to add those options somowhere in tcp transport. 
On other hand, lib is moving right now to 1.0.0, and this may be delayed to 1.1.0

Mliviu79:
Opened #333 which takes this in the shape asked for above — `WriteTimeout` and `ReadTimeout` as fields on `TransportTCP`, reachable through `TransportsConfig`, no package globals. Zero is the default and arms no deadline, so it is a no-op for existing users.

Credit to @j-licht for finding this and for the original diff. Happy to close #333 if you would rather carry your own.

emiago:
The commit references #268 and #333. It has not been pushed. Unrelated untracked files remain untouched.

  Copy-pastable PR/issue comment:

  Thanks @j-licht for reporting and initially addressing this in #268, and @Mliviu79 for the follow-up work in #333.

  I revisited the problem and prefer exposing this through the existing `TransportsConfig` mechanism, without package
  globals or another transport-layer option.

  The implementation adds `ReadTimeout` and `WriteTimeout` to both `TransportTCP` and `TransportWS`. TLS and WSS use
  the same behavior through their embedded TCP and WebSocket transports.

  Example for TCP:

  ```go
  sip.WithTransportLayerTransports(sip.TransportsConfig{
      TCP: &sip.TransportTCP{
          ReadTimeout:  30 * time.Second,
          WriteTimeout: 3 * time.Second,
      },
  })

PR #269, opened 2025-12-01
transport_tcp: reduce TransportBuffer Size to not overload parser

j-licht:
Reduce the TransportBufferReadSize in TCP transport to not overflow the ParserStream as it's ParseMaxMessageLength is the same size.
Maybe we should generally reduce the TransportBufferReadSize.


emiago:
fixed on main. We will follow some other practice and limit reading to 32k
Thanks for reporting this

PR #270, opened 2025-12-01
transport_tcp: fix racy connection creation

j-licht:
the get and create connection in transport layer is racy:
https://github.com/emiago/sipgo/blob/7aff7316d377e3091ff97b6fa539d1c83469af5b/sip/transport_layer.go#L421-L434

Because connection can be created between get and create from another call to the same address.
To prevent double connections and then resulting broken stream parsing, add a mutex in transport_tcp and check if connection really not exists before creating a new one.

### DIFF
--- a/sip/transport_tcp.go
+++ b/sip/transport_tcp.go
@@ -17,6 +17,7 @@ type TransportTCP struct {
 	transport       string
 	parser          *Parser
 	log             *slog.Logger
+	createMu        sync.Mutex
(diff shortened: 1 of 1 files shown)


emiago:
Is this something you can confirm or it is your assumption? This problem was resolved on transport level?

emiago:
Hi @j-licht . Please check main branch and close issue if resolved. It was solved on connection pool.

PR #272, opened 2025-12-08
Support for tel uri

basicus:
Added support for the tel scheme in Uri

### DIFF
--- a/client.go
+++ b/client.go
@@ -539,6 +539,14 @@ func ClientRequestIncreaseCSEQ(c *Client, req *sip.Request) error {
 	return nil
 }
 
+func ClientRequestForceLocalAddr(c *Client, req *sip.Request) error {
+	// Set request local address if it defined
+	if c.connAddr.IP != nil {
+		c.connAddr.Copy(&req.Laddr)
+	}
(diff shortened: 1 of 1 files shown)


basicus:
Hi! 
@emiago 
I used RFC https://datatracker.ietf.org/doc/html/rfc3966 This is main actual tel uri standard. Other rfc is describing specific additional params.

tel uri contain info about subscriber (telephone-subscriber field in ABNF) like username in sip uri. 
Specifically in RFC said:

> The "tel" URI is a globally unique identifier ("name") only; it does
>    not describe the steps necessary to reach a particular number and
>    does not imply dialling semantics.  Furthermore, it does not refer to
>    a specific physical device, only to a telephone number.
> 

telephone-subscriber can be global-number / local-number and contain like SIP uri-params.  Its contain info about user and not contain info about host. 

I've modified Uri parsing flow by skipping some steps (like uriStateHost, uriStatePort) because they are not in the scheme Tel uri. I also added the build of a valid string, including tests.


With this fix made possible correct parse VoLTE SIP INVTE with tel uri in From/To Fields and in Request Line.

PR #275, opened 2025-12-12
use DefaultLogger() everywhere a logger is needed in sip package

jwetzell:
all logging under the `sip` package should use the `DefaultLogger()` method when needing a logger instance.


emiago:
nice. Merging

PR #282, opened 2026-01-19
Make header and URI parameters ordered

dennwc:
This changes switches from a map representation of header/URI params to a slice.

What changes:

- Parameters now preserve the order in which they appear in the original message.
- Parameters won't be shuffled around during encoding of SIP messages, which helps with testing.
- Zero value of `HeaderParams` is now safe to use, it doesn't require `NewParams()`.
- This change improves performance and reduces allocations (see below).
- Some minor changes needs to be done to update existing code (if it uses underlying maps directly).

## Benchmarking results

~25-30% faster header parsing, ~50-60% less allocations when parsing headers with params.

```
                               │   old.txt   │               new.txt                │
                               │   sec/op    │    sec/op     vs base                │
HeaderParams/MAP-12              160.0n ± 1%   112.2n ±  1%  -29.87% (p=0.000 n=10)
ParserHeaders/ViaHeader-12       191.6n ± 2%   138.0n ±  1%  -27.99% (p=0.000 n=10)
ParserHeaders/ToHeader-12        154.0n ± 1%   110.7n ±  1%  -28.14% (p=0.000 n=10)
ParserHeaders/FromHeader-12      153.5n ± 0%   109.9n ±  0%  -28.35% (p=0.000 n=10)
ParserHeaders/ContactHeader-12   177.3n ± 1%   136.1n ±  1%  -23.22% (p=0.000 n=10)
ParserHeaders/Route-12           295.5n ± 1%   247.1n ±  1%  -16.40% (p=0.001 n=10)

                               │    old.txt     │                 new.txt                 │
                               │      B/op      │     B/op      vs base                   │
HeaderParams/MAP-12                85.00 ± 1%      208.00 ± 0%  +144.71% (p=0.000 n=10)
ParserHeaders/ViaHeader-12         464.0 ± 0%       128.0 ± 0%   -72.41% (p=0.000 n=10)
ParserHeaders/ToHeader-12          304.0 ± 0%       208.0 ± 0%   -31.58% (p=0.000 n=10)
ParserHeaders/FromHeader-12        304.0 ± 0%       208.0 ± 0%   -31.58% (p=0.000 n=10)
ParserHeaders/ContactHeader-12     304.0 ± 0%       208.0 ± 0%   -31.58% (p=0.000 n=10)
ParserHeaders/Route-12             464.0 ± 0%       288.0 ± 0%   -37.93% (p=0.000 n=10)

                               │   old.txt    │               new.txt                │
                               │  allocs/op   │ allocs/op   vs base                  │
HeaderParams/MAP-12              3.000 ± 0%     4.000 ± 0%  +33.33% (p=0.000 n=10)
ParserHeaders/ViaHeader-12       4.000 ± 0%     2.000 ± 0%  -50.00% (p=0.000 n=10)
ParserHeaders/ToHeader-12        5.000 ± 0%     2.000 ± 0%  -60.00% (p=0.000 n=10)
ParserHeaders/FromHeader-12      5.000 ± 0%     2.000 ± 0%  -60.00% (p=0.000 n=10)
ParserHeaders/ContactHeader-12   5.000 ± 0%     2.000 ± 0%  -60.00% (p=0.000 n=10)
ParserHeaders/Route-12           4.000 ± 0%     3.000 ± 0%  -25.00% (p=0.000 n=10)
```

The `HeaderParams/MAP` shows the same performance improvement, but a bit more allocations. I believe this is caused by `NewParams()` allocating a bit more space when necessary to compensate for the real-world usage.

Typically, the number of parameters is limited by 4-5, which doesn't justify the overhead of Go maps, so these numbers make sense for the real SIP headers.

## Compatibility with existing code

Since the underlying type of `HeaderParams` is different, there are a couple of changes to the client code that must be done:

- If the code only uses existing functions (`NewParams`, `Get`, `Add`), then **no changes** are necessary!
- Initialization must be changed from `{"a":"1", "b":"2"}` to `{{"a","1"}, {"b":"2"}}`.
- Direct lookups via `params[k]` must be replaced with `params.Get(k)` or `params.GetOr(k, "")` (second one omits bool return).

There are some **side affects** that must be considered:

- `HeaderParams` is no longer a reference type, so passing it to functions and expecting changes to propagate requires an extra pointer.
- Methods like `Add` and `Remove` now require a pointer receiver. This should not require changes, since map was always a reference.

### DIFF
--- a/sip/header_params.go
+++ b/sip/header_params.go
@@ -2,107 +2,133 @@ package sip
 
 import (
 	"io"
+	"slices"
 	"strings"
 )
 
+// HeaderKV is a key-value pair for URI or header params.
 type HeaderKV struct {
 	K string
 	V string
 }
 
-// HeaderParams are key value params. They do not provide order by default due to performance reasons
-type HeaderParams map[string]string
+// HeaderParams are key value params.
+type HeaderParams []HeaderKV
 
-// Create an empty set of parameters.
+// NewParams creates an empty set of parameters.
 func NewParams() HeaderParams {
-	return HeaderParams{}
+	// Typical number of params:
+	// URI: 1-2
+	// Via: 1-2
+	// Route: 4
+	return make(HeaderParams, 0, 4)
 }
 
 // Items returns the entire parameter map.
 func (hp HeaderParams) Items() map[string]string {
 	m := make(map[string]string, len(hp))
-	for k, v := range hp {
-		m[k] = v
+	for _, kv := range hp {
+		m[kv.K] = kv.V
 	}
 	return m
 }
 
-// Keys return a slice of keys, in order.
+// Keys return a slice of keys, in order of appearance.
 func (hp HeaderParams) Keys() []string {
-	s := make([]string, len(hp))
-	i := 0
-	for k := range hp {
-		s[i] = k
-		i++
+	s := make([]string, 0, len(hp))
+	for _, kv := range hp {
+		if slices.Contains(s, kv.K) {
+			continue
+		}
+		s = append(s, kv.K)
 	}
 	return s
 }
 
-// Get returns existing key
+func (hp HeaderParams) index(key string) int {
+	for i, kv := range hp {
+		if kv.K == key {
+			return i
+		}
+	}
+	return -1
+}
+
+// Get returns a value for a given key, if it exists.
 func (hp HeaderParams) Get(key string) (string, bool) {
-	v, ok := hp[key]
-	return v, ok
+	for _, kv := range hp {
+		if kv.K == key {
+			return kv.V, true
+		}
+	}
+	return "", false
 }
 
-// Add will add new key:val. If key exists it will be overwriten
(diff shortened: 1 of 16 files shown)


sip/header_params.go
```diff
@@ -2,139 +2,178 @@ package sip
...
+	if i := hp.index(key); i >= 0 {
+		(*hp)[i].V = val
+	} else {
+		hp.Append(key, val)
+	}
+}
+
+// Append a new key-value pair. It doesn't check if it exists already.
+// This is a more efficient way to add a missing attribute.
+func (hp *HeaderParams) Append(key string, val string) {
```
emiago: Based on RFC
>  Even though an arbitrary number of URI parameters may be
         included in a URI, any given parameter-name MUST NOT appear
         more than once.

I would like to avoid this API as exposed. 
Generally there is small gain here as key value pairs are in very low number. 

If there is any need for perfomance gain, API for multiple key values should be created, so that allocs behind happen
once. 
This kind of apis are more enforcing multiple call to append!

PR #284, opened 2026-01-30
Fix ACK request address handling to preserve connection information

pablodz:
- Add Destination, Laddr to newAckRequestUAC for 2xx ACK responses
- Add Destination, Laddr, and raddr to newAckRequestNon2xx for non-2xx ACK responses
- Ensures ACK requests maintain same connection properties as original INVITE
- Fixes routing issues where ACK could be sent to wrong destination


emiago:
Ok. I merged but I added fix after. `SetDestination` we will avoid from now on. It is racy and it should be only used on transport layer. Building request should never use this anymore. Thnx for other fixing. I left `Laddr` changes.

PR #288, opened 2026-03-09
Notify pending transactions on TCP/TLS connection close

Odinvt:
**Scenario :**

When a TCP/TLS connection is closed by the remote end, readConnection exits without notifying the transaction layer. Pending client transactions on that connection wait the full Timer_B (32s) before timing out.

**Fix :**

This adds a callback from the transport to the transaction layer on connection close. Matching client transactions receive client_input_transport_err, which is already handled in every FSM state. The goroutine pattern matches existing transport error handling in ack()/resend().

**How this was found :**

During mTLS testing against Kamailio with require_certificate=yes, a client connecting with an invalid certificate completes the TLS handshake (TLS 1.2 sends client cert after ServerHello) but the server immediately closes the connection upon evaluating the cert. Our REGISTER transaction was left hanging for 32 seconds despite the connection being dead.

### DIFF
--- a/sip/transaction_layer.go
+++ b/sip/transaction_layer.go
@@ -49,8 +49,8 @@ func WithTransactionLayerUnhandledResponseHandler(f func(r *Response)) Transacti
 }
 
 // WithTransactionLayerTerminateOnConnClose enables termination of pending
-// client transactions when the underlying connection closes, instead of
-// waiting for Timer_B to expire.
+// client and server transactions when the underlying connection closes,
+// instead of waiting for timers to expire.
 //
 // Experimental
 func WithTransactionLayerTerminateOnConnClose() TransactionLayerOption {
@@ -103,6 +103,7 @@ func (txl *TransactionLayer) OnRequest(h TransactionRequestHandler) {
 func (txl *TransactionLayer) OnConnectionClose(conn Connection) {
 	if txl.terminateOnConnClose {
 		txl.terminateClientTransactions(conn)
+		txl.terminateServerTransactions(conn)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Odinvt:
> I would use global but ok it can be tranport layer parameter.

Sorry for not looking deeper into the global var pattern before submitting, I should have checked your conventions first.

> I am just wondering why would we not include server transactions as well.

Our use case is UAC-heavy (SIP load testing), so client transactions are where I hit this in practice. I haven't faced the server tx side yet, which is why it wasn't included. But looking at the code, here's what I found (please do correct me if I'm wrong):

**Server tx lifecycle relative to `readConnection`:**

- `readConnection` goroutine (per conn)
  - `conn.Read(buf)` → `parseStream(data)` → `handler(msg)` (= `txl.handleMessage`)
    - `go handleRequestBackground(req)` — NEW goroutine, detached from read loop
      - `serverTxCreate` → `GetConnection(source)` — retrieves SAME conn from pool
      - `NewServerTx(key, req, conn)` — tx holds that conn reference
      - `txl.reqHandler(req, tx)` — app handler runs in this goroutine

The `go` in `handleMessage` means the read loop doesn't block on server tx creation or the app handler. They run detached.

**When connection dies:**

1. `readConnection` gets EOF
2. Defers fire LIFO: `onConnClose(conn)` first, then `pool.CloseAndDelete(conn, raddr)`
3. The server tx still holds the same `conn` pointer
4. Next `tx.Respond()` call hits `conn.WriteMsg()` on a closed conn, which returns `server_input_transport_err`, FSM handles it and transitions to Terminated

So server transactions already self-terminate on the next write attempt. The difference from client transactions: there's no Timer_B equivalent sitting idle for 32s. The server tx lifecycle is bounded by application code duration (handler runs, calls `Respond()`, fails, done). And `TerminateGracefully()` runs after the handler returns as a safety net.

**The concern with including server tx:** firing `server_input_transport_err` on connection close would race with the application handler goroutine that's still running. The handler might be mid-work (DB queries, external calls) before calling `tx.Respond()`. If I terminate the tx under it, the handler sees an unexpected terminated state. That's a behavioral change for application code.

It could still make sense as a separate opt-in, but it needs its own analysis since the failure mode is different: client tx = stuck waiting for a response that will never come, server tx = app handler racing with early termination. Happy to look into it as a follow-up if you think it's worth exploring.

Odinvt:
Actually thinking about this more, you're right. `OnConnectionClose` is on the transaction layer, not on individual transactions. Applications can't hook into it to clean up their server tx handlers.

Server transactions already have `tx.Done()` and `tx.Err()` as notification channels, and the FSM handles `server_input_transport_err` in every state. Terminating server tx on connection close would work: applications that select on `tx.Done()` get notified immediately, applications that don't just hit the write error on `Respond()` as before, which is harmless since the tx is already terminated by then. The `fsmMu` lock prevents the race between termination and `Respond()`.

Up to you if you want me to add server transactions to the same flag in this PR or keep it as a follow-up.

PR #294, opened 2026-04-11
sip: ignore UDP probe noise from LB health checks

hujm2023:


emiago:
Should be possible todo after adding this #295 , and maybe if you give more context we can have it as part of lib as well.

emiago:
Hi @hujm2023 could we now use this as predefinted 
`TransportReadFilter` . 

Consider that above needs to work even for streaming network and not just datagrams. 
I am not aware of this PING checks or why your LB does not use something as SIP standard, but it would be good not to close this if more users have same issue, so some helper functions could be provided.

PR #296, opened 2026-04-23
sip: send stateless 400 Bad Request for malformed requests

MohamadAbdelmoniem:
## Problem

When the transaction layer receives a request with missing mandatory headers (e.g. `CSeq`), `makeServerTxKey` fails and the request is silently dropped. The sender never receives a response and retransmits indefinitely, flooding server logs with errors like:

    Server tx failed to handle request: make key failed: 'CSeq' header not found

## Fix

Per RFC 3261 §16.3, a proxy that receives a request with a malformed required header SHOULD return a 400 Bad Request response.

Adds a `rejectMalformedRequest` helper that sends a stateless 400 Bad Request via the transport layer when transaction key creation fails. This covers both regular requests and CANCEL matching.

`NewResponseFromRequest` already handles nil CSeq, From, To, and Call-ID gracefully by skipping them, so the response is safe to build from a partially malformed request.

## Test

Integration test `TestTransactionLayerMalformedRequestStateless400` sends a REGISTER missing CSeq over UDP and verifies that a 400 Bad Request response is received instead of silence.

### DIFF
--- a/sip/transaction_layer.go
+++ b/sip/transaction_layer.go
@@ -215,7 +215,6 @@ func (txl *TransactionLayer) rejectMalformedRequest(req *Request, reason error) 
 	// Build a minimal 400 response from whatever headers the request has.
 	// NewResponseFromRequest safely skips nil CSeq, From, To, Call-ID.
 	res := NewResponseFromRequest(req, StatusBadRequest, "Bad Request", nil)
-	res.SetDestination(req.Source())
 
 	if err := txl.tpl.WriteMsg(res); err != nil {
 		txl.log.Error("Failed to send stateless 400 for malformed request",
--- a/sip/transaction_layer.go
+++ b/sip/transaction_layer.go
@@ -162,6 +162,7 @@ func (txl *TransactionLayer) handleRequest(req *Request) error {
 		// For now we only match INVITE
 		key, err := makeServerTxKey(req, INVITE)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


noureldenn:
i have faced this problem also looding server logs with errors like:

Server tx failed to handle request: make key failed: 'CSeq' header not found

sip/transaction_layer.go
```diff
@@ -186,12 +187,45 @@ func (txl *TransactionLayer) handleRequest(req *Request) error {
...
+//	If the request contains a malformed header field that the proxy
+//	requires, the proxy SHOULD return a 400 (Bad Request) response.
+//
+// Without this, the sender retransmits indefinitely because it never
+// receives any response.
+func (txl *TransactionLayer) rejectMalformedRequest(req *Request, reason error) {
+	// Build a minimal 400 response from whatever headers the request has.
+	// NewResponseFromRequest safely skips nil CSeq, From, To, Call-ID.
+	res := NewResponseFromRequest(req, StatusBadRequest, "Bad Request", nil)
+	res.SetDestination(req.Source())
```
emiago: Pls remove this. It is handled by above func more correctly

emiago:
LGTM. Just one comment above.

sip/transaction_layer.go
MohamadAbdelmoniem: Done

Issue #226, opened 2025-05-01
Failing to parse SIP URI with IPv6 in contact header


darkrift:
When a contact header has an IPv6 in it, the header fails to be parsed and treats a portion of the IPv6 as the port. The current issue is not specific to the contact header.

Reproducer :
```
package sip

import (
	"testing"

	"github.com/stretchr/testify/require"
)

func TestParseContactWithIPv6Value(t *testing.T) {
	_, err := headerParserContact("Contact", "Contact: \"my name\" <sip:blablabla@[64:ff9b::c0a8:7f65]:59370>")
	require.NoError(t, err)
}

```

```
=== RUN   TestParseContactWithIPv6Value
    parse_contact_test.go:11: 
        	Error Trace:	/Users/rlavoie/go/src/github.com/emiago/sipgo/sip/parse_contact_test.go:11
        	Error:      	Received unexpected error:
        	            	strconv.Atoi: parsing "ff9b::c0a8:7f65]:59370": invalid syntax
        	Test:       	TestParseContactWithIPv6Value
--- FAIL: TestParseContactWithIPv6Value (0.00s)

FAIL
```

The culprit is `uriStateHost` in `parse_uri.go` that clashes parsing the semi-colon. RFC3261 defines a host with IPv6 to be enclosed within `[` and `]` that is not taken into account here:
```
hostport         =  host [ ":" port ]
host             =  hostname / IPv4address / IPv6reference
hostname         =  *( domainlabel "." ) toplabel [ "." ]
IPv4address    =  1*3DIGIT "." 1*3DIGIT "." 1*3DIGIT "." 1*3DIGIT
IPv6reference  =  "[" IPv6address "]"
IPv6address    =  hexpart [ ":" IPv4address ]
hexpart        =  hexseq / hexseq "::" [ hexseq ] / "::" [ hexseq ]
hexseq         =  hex4 *( ":" hex4)
hex4           =  1*4HEXDIG
port           =  1*DIGIT
```

A minimum working version could be the following but this won't be sufficient to marshal back the header to text in order to enclose the IPv6 within square brackets.
```
func uriStateHost(uri *Uri, s string) (uriFSM, string, error) {
	var inIPv6 bool
	var hasIPv6 bool
	for i, c := range s {
		if i == 0 && c == '[' {
			inIPv6 = true
		} else if c == ']' && inIPv6 {
			inIPv6 = false
			hasIPv6 = true
		}

		if c == ':' && !inIPv6 {
[...]

PR #300, opened 2026-05-02
fix(parse_via): handle IPv6 host literals in Via header parsing

Booyaka101:
## What this fixes

`viaStateHost` in `sip/parse_via.go` previously walked all colons in a Via header host greedily, without IPv6 bracket awareness. For a header like:

```
Via: SIP/2.0/TCP [2001:67c:b0c:5::192];branch=z9hG4bKB5mrcD1BXK68a
```

the loop ends with `colonInd` pointing at the second `:` of `::`. `strconv.Atoi("192]")` then fails, the FSM bails out, and the SIP message is silently dropped at the transport layer.

## How it was diagnosed

Surfaced downstream as [livekit/sip#640](https://github.com/livekit/sip/issues/640) — IPv6 inbound INVITEs from FreeSWITCH (both UDP and TCP transports) failing with `strconv.Atoi: parsing "67c:b0c:5::192]:5060": invalid syntax: EOF on reading line`. Original report attributed the bug to livekit-sip's wrapping code, but tracing through the FSM character-by-character points cleanly at this function: livekit-sip's own `strconv.Atoi` calls are all in non-IPv6-relevant paths.

## Precedent

This is the same shape of bug that `parse_uri.go` had until [#226](https://github.com/emiago/sipgo/issues/226) was fixed in May 2025 by adding `uriStateHostIPV6` (`parse_uri.go:117`). The Via parser was missed in that round.

## The fix

Detect the leading `[` up front, capture everything up to the matching `]` as the host (stored without brackets to match `uriIP()`'s expectation at serialization — see `sip/utils.go:405`), then dispatch on whatever follows the bracket:

- `:port` — parse port via `strconv.Atoi` on the substring up to the next `;` or end-of-string
- `;params` — hand off to `viaStateParams` directly
- end-of-string — return cleanly with no params

Falls back to the original colon-walking path for non-IPv6 hosts unchanged. No regression risk for IPv4 / hostname callers.

## Diff shape

```
 sip/parse_via.go   | +46
 sip/parser_test.go | +47
```

All net additions. The only change to `parse_via.go:viaStateHost` is a guard at the top that handles the IPv6 bracket case before the existing colon-walking logic; non-IPv6 inputs hit the original code path unmodified.

## Test coverage

Added an `IPv6` subtest under `TestParseHeaders/ViaHeader` covering three real-world shapes:

| Case | Header | Expected |
|---|---|---|
| `with_port` | `[2001:67c:b0c:5::192]:5060;branch=...` | host=`2001:67c:b0c:5::192`, port=5060 |
| `no_port` | `[2001:67c:b0c:5::192];branch=...` (the livekit/sip#640 case) | host=`2001:67c:b0c:5::192`, port=0 |
| `loopback` | `[::1]:5060;branch=...` | host=`::1`, port=5060 |

Without the fix, the `no_port` case fails (`host=""`); with the fix, all three pass.

## Verification

- `go test ./sip/ -run "TestParseHeaders" -count=1` — **PASS** (all subtests including the new `IPv6` block)
- `go test ./sip/ -count=1` — **PASS** (full sip package: parser, headers, transactions, fuzz, stream)
- I deliberately reverted the fix and re-ran to confirm `no_port` fails on the unfixed parser before applying — empirical regression-test discipline

The integration test failure at the repo root (`server_integration_test.go:36: testdata/certs/client.crt: no matching files found`) is pre-existing — `//go:embed` fails on missing test certs that aren't checked into the repo. Unrelated to this PR.

## Credit

Original report by @arsenieciprian on [livekit/sip#640](https://github.com/livekit/sip/issues/640). They did the empirical work to demonstrate the bug across UDP and TCP transports — the FSM trace and fix shape here is downstream of that report.

### DIFF
--- a/sip/parse_via.go
+++ b/sip/parse_via.go
@@ -71,6 +71,52 @@ func viaStateProtocolTransport(h *ViaHeader, s string) (viaFSM, int, error) {
 }
 
 func viaStateHost(h *ViaHeader, s string) (viaFSM, int, error) {
+	// IPv6 host literal: "[addr]" with optional ":port" and ";params".
+	// The colons inside the IPv6 literal must not be confused with the
+	// host:port separator, so this case is handled before the generic
+	// colon-walking path below.
+	if len(s) > 0 && s[0] == '[' {
+		closeIdx := strings.Index(s, "]")
+		if closeIdx < 0 {
+			return nil, 0, errors.New("Malformed IPv6 host in Via header: no closing bracket")
+		}
+		// Store the bare IPv6 address without brackets to match the
+		// convention expected by uriIP() at serialization time.
+		h.Host = strings.TrimSpace(s[1:closeIdx])
+
+		rest := s[closeIdx+1:]
+		baseOff := closeIdx + 1
+		if len(rest) == 0 {
+			return nil, 0, nil
+		}
+		if rest[0] == ':' {
+			// Optional port follows the IPv6 literal.
+			endIdx := strings.IndexByte(rest, ';')
+			var portStr string
+			if endIdx < 0 {
+				portStr = rest[1:]
+			} else {
+				portStr = rest[1:endIdx]
+			}
+			port, err := strconv.Atoi(strings.TrimSpace(portStr))
+			if err != nil {
+				return nil, 0, nil
+			}
+			h.Port = port
+			if endIdx < 0 {
+				return nil, 0, nil
+			}
+			return viaStateParams, baseOff + endIdx + 1, nil
+		}
+		if rest[0] == ';' {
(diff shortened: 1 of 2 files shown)


emiago:
Nice that IPV6 getting some attention.

thetillhoff:
Amazing that this was already fixed and merged!
Do you have an estimate when it will be included in a release? I'm currently waiting for it haha

thetillhoff:
@emiago do you have an update on when this ipv6 fix will be released aka when the next version will be released?

emiago:
@thetillhoff will do this week. :)

Issue #295, opened 2026-04-11
Feature: Add message prefiltering on read


emiago:
## Tickets
Give user possibility to prefilter messages on `Read`. Here are related issues 
- #294 
- #292 

## Solution proposal
Hook on transport layer or per specific listener. 
Specific listener would give user to hook only when it needs to because it can have trusted/untrusted networks.

Args:
- message as byte array
- remote and local addr or just connection interface

Return
- int  - how many bytes to use after filtering. 0 would mean message should be skipped. Forces user to reuse buffer
- error - In case error message processing should be stopped and read connection will bes 

## Goal 
Net message filtering or catching corner incompatible messages before hitting parser

yangjuncode:
would it be better return new filtered byte array as result and err, so user can patch the income if they want

emiago:
@yangjuncode 
There can be use case that filtered is larger output, so yes that could make sense. I wanted somehow to force user to reuse buffer, but it would be both possible.
Thnaks for feedback.

emiago:
ok added now support as single global option for all transports by passing 
`WithTransportLayearReadFilter`  . I have settled on this 

`type TransportReadFilter func(info TransportReadProps, data []byte) ([]byte, error)` 

Some notes:
- it is just a simple hook that applies to all network. User should use laddr or remote addr to determine context or interface
- TCP or other stream filtering is something user needs to handle more carefully so also transport as data is passed 

I hope this is enough to allow users to fix their corner cases or even used to detect bad message or even IP filtering.

Give feedback on this

Issue #251, opened 2025-08-12
SIP headers with folded lines fail to parse.


DentonGentry:
Below is a SIP message where the WWW-Authenticate header uses line folding.
It was received from the AT&T Wireless VoWifi server — so, a production service.

```
SIP/2.0 401 Unauthorized\r\n
Call-ID: 01234567\r\n
Via: SIP/2.0/TCP [::1]:5060;received=::1;branch=z9hG4bK.1;rport=48512\r\n
To: <sip:30123456789@one.att.net>;tag=abcdef\r\n
From: \"30123456789\" <sip:30123456789@one.att.net>;tag=abcdef\r\n
CSeq: 1 REGISTER\r\n
Date: Tue, 12 Aug 2025 01:20:52 GMT\r\n
Server: Alcatel-Lucent-HPSS/3.0.3\r\n
WWW-Authenticate: Digest realm=\"one.att.net\",\r\n
   nonce=\"abcdefghijklmnopqrstuvwxyz=\",\r\n
   opaque=\"ALU:abcdefghijklmnopqrstuvwxyz__\",\r\n
   algorithm=AKAv1-MD5,\r\n
   qop=\"auth\"\r\n
Content-Length: 0\r\n
\r\n
```

Attempting to parse this results in an error, `ERROR failed to parse caller=TransportLayer caller=Transport<TCP> error="field name with no value in header:    nonce=\"abcdefghijklmnopqrstuvwxyz=\",: EOF on reading line"`

I've constructed a test case to demonstrate the failure in https://github.com/DentonGentry/sipgo/tree/dentongentry-line-folding which fails. I haven't developed a fix, just demonstrated the issue in a form which might turn into a test case for a fix.

emiago:
Hi @DentonGentry . Welcome.  Folding lines not (yet) supported, although it would nice that SIP followed HTTP in this and removed this behavior. Can I ask is this some legacy system?

emiago:
We should try support folding by using this #295 as quick solution.

PR #301, opened 2026-05-02
fix(parser): unfold continuation header lines per RFC 3261 §7.3.1 (#251)

Booyaka101:
Refs #251.

## Context — and why this is filed as a draft

@emiago — you mentioned on April 13 that you'd prefer to handle line folding via #295's prefilter feature. Filing this as a draft so you have a turnkey alternative to consider, but happy to step aside immediately if you'd rather proceed with the prefilter route — please just close this and I won't take it personally. The prefilter feature serves a broader set of needs than this single bug, so the two paths aren't mutually exclusive even.

The reason I'm offering the parser-level fix as an option: RFC 3261 §7.3.1 explicitly requires support for header continuation, so the current behavior is non-compliant by default. Pushing the work to every sipgo user via a prefilter hook means each consumer reimplements the same RFC normalization, and silently dropped messages remain the default for users who haven't yet realized line folding can occur on their carrier.

## What this fixes

@DentonGentry's [#251](https://github.com/emiago/sipgo/issues/251) reproduces the failure with a real `WWW-Authenticate` header from the AT&T VoWiFi P-CSCF (the production service handling US AT&T Wireless voice and SMS). Folded over four continuation lines:

```
WWW-Authenticate: Digest realm="one.att.net",
   nonce="abcdefghijklmnopqrstuvwxyz=",
   opaque="ALU:abcdefghijklmnopqrstuvwxyz__",
   algorithm=AKAv1-MD5,
   qop="auth"
```

The current parser treats line two onward as fresh headers and surfaces `field name with no value in header: "   nonce=..."`, dropping the message entirely.

## How it works

`parseNextHeader` now sweeps continuation lines into the logical header line before handing it to the per-header parser. The fix is wholly contained in `parseNextHeader`; `nextLine`, the start-line parser, and the per-header parsers are untouched.

```go
// RFC 3261 §7.3.1: a header field may be continued onto the next line if
// that line begins with at least one SP or HT. The CRLF and surrounding
// whitespace must be replaced by a single SP.
for n < len(data) && (data[n] == ' ' || data[n] == '\t') {
    cont, contN, contErr := nextLine(data[n:])
    if contErr != nil {
        if contErr == io.EOF {
            return out, 0, io.ErrUnexpectedEOF
        }
        return out, 0, contErr
    }
    folded := make([]byte, 0, len(line)+1+len(cont))
    folded = append(folded, line...)
    folded = append(folded, ' ')
    i := 0
    for i < len(cont) && (cont[i] == ' ' || cont[i] == '\t') {
        i++
    }
    folded = append(folded, cont[i:]...)
    line = folded
    n += contN
}
```

## Properties

- **End-of-headers boundary is unaffected.** The empty line that terminates the header section starts with `\r`, which fails the SP/HT check, so it short-circuits to `errParseNoMoreHeaders` exactly as before.
- **Streaming partial-data contract preserved.** A truncated continuation surfaces as `io.ErrUnexpectedEOF` with `n=0`, matching the existing semantics for an incomplete first line, so the stream parser retries with more bytes.
- **`nextLine` itself unchanged.** The start-line path (where folding must NOT apply) is untouched.
- **Zero allocation on the no-fold path.** Non-folded headers (the common case) take the original code path; the new slice is allocated only when at least one continuation exists.
- **N-line continuations supported.** The AT&T trace folds across four continuation lines; the loop handles arbitrary depth.
- **Strict behavior superset.** Any message that parsed before still parses identically; only previously-erroring messages now succeed.

## Tests

`sip/parser_folded_test.go` adds two test functions:

- `TestParseFoldedHeaders_AT_T_Repro` — replays @DentonGentry's full `401 Unauthorized` trace and asserts every `WWW-Authenticate` subfield survives the fold.
- `TestParseFoldedHeaders_Cases` — isolates the folding behavior on a generic header (`Server`, no custom parser) with three sub-cases: spaces-prefixed continuation, tab-prefixed continuation, and a no-folding control.

## Verification

| Platform | Pre-fix | Post-fix |
|---|---|---|
| Windows (Go 1.26.2 native) | folding tests fail with the exact error from #251; full suite otherwise green | all four new tests pass; full sip-package suite green |
| Linux (`golang:1.26` Docker, amd64) | folding tests fail with the same error | all four new tests pass; full sip-package suite green |

## Credit

Original report and the test trace by @DentonGentry in #251. They isolated this against the AT&T VoWiFi P-CSCF in production and even pushed a demo branch reproducing the failure — saving the diagnosis work entirely.

### DIFF
--- a/sip/parser.go
+++ b/sip/parser.go
@@ -151,32 +151,45 @@ func (p *Parser) parseNextHeader(out []Header, data []byte) ([]Header, int, erro
 
 	// RFC 3261 §7.3.1: a header field may be continued onto the next line if
 	// that line begins with at least one SP or HT. The CRLF and surrounding
-	// whitespace must be replaced by a single SP. Sweep up any continuation
-	// lines now so the downstream header parser sees a single logical line.
-	for n < len(data) && (data[n] == ' ' || data[n] == '\t') {
-		cont, contN, contErr := nextLine(data[n:])
-		if contErr != nil {
-			// Need more bytes / malformed CRLF on the continuation. Surface
-			// the same way an incomplete first line would surface so the
-			// stream parser can retry with more data.
-			if contErr == io.EOF {
-				return out, 0, io.ErrUnexpectedEOF
+	// whitespace must be replaced by a single SP. Pre-scan the continuation
+	// lines so we can size the output exactly and fold with a single
+	// allocation regardless of how many continuations there are.
+	if n < len(data) && (data[n] == ' ' || data[n] == '\t') {
+		type foldSegment struct{ start, end int }
+		var segs []foldSegment
+		totalSize := len(line)
+		for n < len(data) && (data[n] == ' ' || data[n] == '\t') {
+			cont, contN, contErr := nextLine(data[n:])
+			if contErr != nil {
+				// Need more bytes / malformed CRLF on the continuation.
+				// Surface the same way an incomplete first line would, so the
+				// stream parser can retry with more data.
+				if contErr == io.EOF {
+					return out, 0, io.ErrUnexpectedEOF
+				}
+				return out, 0, contErr
 			}
-			return out, 0, contErr
+			// Trim leading SP/HT runs from the continuation per RFC 3261 §7.3.1.
+			contStart := n
+			contEnd := n + len(cont)
+			for contStart < contEnd && (data[contStart] == ' ' || data[contStart] == '\t') {
+				contStart++
+			}
+			segs = append(segs, foldSegment{contStart, contEnd})
+			totalSize += 1 + (contEnd - contStart) // +1 for the SP separator
+			n += contN
 		}
-		// `line` may alias the original buffer; allocate a fresh slice the
-		// first time we fold so we don't write into the caller's data.
-		folded := make([]byte, 0, len(line)+1+len(cont))
+
+		// Single allocation, exactly sized, regardless of continuation count.
+		// `line` may alias the original buffer; copying into a fresh slice
+		// also keeps us from writing into the caller's data.
+		folded := make([]byte, 0, totalSize)
 		folded = append(folded, line...)
-		folded = append(folded, ' ')
-		// Trim leading SP/HT runs from the continuation per RFC 3261 §7.3.1.
-		i := 0
-		for i < len(cont) && (cont[i] == ' ' || cont[i] == '\t') {
-			i++
+		for _, s := range segs {
+			folded = append(folded, ' ')
+			folded = append(folded, data[s.start:s.end]...)
 		}
-		folded = append(folded, cont[i:]...)
 		line = folded
-		n += contN
 	}
 
 	out, err = p.headersParsers.ParseHeader(out, line)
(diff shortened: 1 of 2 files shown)


yangjuncode:
wow, good work. and i agree with you that cause RFC 3261 §7.3.1 state it, it should be better fix in parser than in prefilter.

emiago:
Hi @Booyaka101 I like the patch, but I do not like this double allocation. 
In other words I would leave header multiline as it is, or parser of header could/should ignore extra spaces when getting parsed. As it is valid to consume it is valid also to be pushed. You may notice that parser is not anyway doing here extra except I think triming front and back.

Maybe I am overlooking something, but I see this extra thing here. Pls fill in in case I am wrong

Booyaka101:
Thanks for the review.

You're right about the allocation — for a header with N continuation lines (the AT&T \`WWW-Authenticate\` trace has 4) the original sweep was O(N) allocations and O(N²) bytes copied because each iteration grew \`folded\` from scratch. Pushed [\`788acb7\`](https://github.com/emiago/sipgo/pull/301/commits/788acb74a) which pre-scans the continuations to compute the exact folded size, then allocates the output buffer once. Total bytes copied is now O(N) regardless of continuation count. Suite passes on both Windows and Linux (Docker \`golang:1.26\`).

On leaving the fold-whitespace in the bytes handed to per-header parsers, I'd push back. RFC 3261 §7.3.1 is explicit: *"The CRLF and any subsequent whitespace are NOT considered part of the header field's value. Implementations MUST recognize line folding and treat the folded value as if it were on a single line."* And concretely in the current parser layout:

- [\`parse_header.go:79\`](https://github.com/emiago/sipgo/blob/main/sip/parse_header.go#L79) does \`bytes.TrimSpace(line[colonIdx+1:])\` — that only trims leading and trailing whitespace, so embedded \`\r\n   \` between e.g. \`realm="..."\` and \`nonce="..."\` would survive into the field value.
- For headers without a registered parser (\`WWW-Authenticate\`, \`Authorization\`, \`Server\`, \`Subject\`, etc.) it falls through to \`NewHeader(name, fieldValue)\`, so callers reading \`header.Value()\` would see embedded CRLFs in the string.
- For the structured parsers, \`parseMaxForwardsHeader\` does \`strconv.ParseUint(headerText, 10, 32)\` directly and \`parseContentLengthHeader\` does the same after a TrimSpace — both fail on any internal whitespace. \`parseCallIdHeader\` only trims front/back so an embedded CRLF would land inside the Call-ID. \`parseCSeqHeader\` looks for the first \`abnf\` (\`" \t"\`) byte to split sequence number from method.
- And every per-header parser would need to grow its own fold-handling, with a separate regression test per parser. The lex-layer fold keeps that one place.

\`TestParseFoldedHeaders_Cases\` already exercises this — \`assert.Equal(\"Alcatel-Lucent HPSS/3.0.3\", servers[0].Value())\` would fail if the value retained \`\r\n   \` between the words.

Happy to discuss further if you'd still prefer the parser-side approach — I can sketch what each affected parser would look like.

emiago:
Hi @Booyaka101 after rethinking, I have added fix based on your claude solution. 
As folding is mostly not something that is found in modern stacks I decided to ignore this allocation. 
Thanks for adding this up.

PR #302, opened 2026-05-08
feat(client): add outbound proxy support via Route header

eschmidbauer:
## Summary
- Add `ClientRequestAddRoute(uri)` option that prepends a loose-routing
  `Route` header to a request — the standard RFC 3261 §16.12.1 outbound
  proxy mechanism. `Request.Destination()` and `Request.Transport()`
  already honor the top Route URI, so the request is dispatched to the
  proxy while the Request-URI is left untouched.
- Add a client-wide outbound proxy: `WithClientOutboundProxy(hostPort, transport)`,
  `Client.SetOutboundProxy(hostPort, transport)`, and `Client.OutboundProxy()`.
  When set, every out-of-dialog request built by the client gets the loose
  Route auto-prepended. In-dialog requests already carrying a Route from
  Record-Route are not modified.
- The optional `transport` hint (e.g. `"tcp"`, `"tls"`) is stamped as
  `;transport=` on the Route URI **and** surfaced via `req.SetTransport`
  before Via is built, so Via, `req.Transport()`, and the connection chosen
  by the transport layer all agree with the proxy. An empty transport
  preserves prior behavior — Recipient `;transport=` and any explicit
  `req.SetTransport` still drive Via build.

## Test plan
- [x] `go test ./...` (full suite, race-free)
- [x] `TestClientRequestAddRoute` — per-request option: lr handling, URI cloning, Destination override
- [x] `TestClientOutboundProxy/OptionInjectsRouteOnBuild` — auto-injection on default build path
- [x] `TestClientOutboundProxy/TransportStampedOnRouteAndVia` — explicit transport drives Route param + Via.Transport + req.Transport()
- [x] `TestClientOutboundProxy/EmptyTransportDoesNotStampParam` — empty transport leaves Route URI free of `;transport=`
- [x] `TestClientOutboundProxy/RecipientTransportRespectedWhenProxyTransportEmpty` — Recipient `;transport=tcp` still drives Via when proxy transport is empty
- [x] `TestClientOutboundProxy/RuntimeSetterAndClear` + `RuntimeSwapReflectedOnNextBuild` — runtime set/clear and transport swap observed on next build
- [x] `TestClientOutboundProxy/InvalidHostPortReturnsError` — bad host:port surfaces the parse error
- [x] `TestClientOutboundProxy/DoesNotOverrideExistingRoute` — in-dialog request with existing Route is not stacked

### DIFF
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net"
+	"sync/atomic"
 
 	"github.com/google/uuid"
 	"github.com/icholy/digest"
@@ -30,6 +31,11 @@ type Client struct {
 	log   *slog.Logger
 
 	connAddr sip.Addr
+
+	// outboundProxy, when non-nil, is auto-prepended as a loose Route on
+	// out-of-dialog requests built by this client. Stored atomically so
+	// SetOutboundProxy can race safely against in-flight requests.
+	outboundProxy atomic.Pointer[sip.Uri]
 
 	// TxRequester allows you to use your transaction requester instead default from transaction layer
 	// Useful only for testing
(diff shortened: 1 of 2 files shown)


emiago:
It is just setting route header or SetDestination. I could see some value having this in lib but only as config value for client. Otherwise user can just manipulate route header or bypass resolving by destination if it needs dynamic routing. 

Seems like you are solving problem, but it can be just a helper function or are you missing something?. Pls do not use AI to respond

eschmidbauer:
I had originally implemented this inside my app using a helper function. Then I considered how often I use Outbound Proxy in SIP (almost always). I thought this would be a beneficial feature for others as well as myself so i wouldn't have to use a helper function in all of my setups.

emiago:
I understand, I probably have even more copy paste, just I think we need to be careful more when adding stuff in lib. 
I like `OutboundProxy` helper thing, but just exposing like this goes a bit out of client.

Give some time. I will leave open, in case more people add there inputs.
Thanks for opening this.

eschmidbauer:
That makes sense, and I agree it’s important to be careful about what becomes part of the library surface area.

My reasoning for requesting this be merged is that outbound proxy support is part of the SIP RFCs and is something essentially every production UAC I’ve used exposes in some form. Since this package implements a UAC in `client.go`, the absence of outbound proxy support makes integration more difficult in environments where proxies are standard infrastructure rather than an edge case.

I also think the helper-based approach keeps the impact fairly minimal since it remains optional and doesn’t complicate the default client flow. I added unit tests around the additional client flow as well so the behavior is covered and protected going forward.

PR #304, opened 2026-05-24
Fix parsing multi-value Contact headers

alexbakker:
Previously, the first parsed Contact would get corrupted with parameters of the second Contact.


emiago:
LGTM. Merging

PR #309, opened 2026-06-04
fix(transaction): release connection reference when a server transaction terminates

simonrakcsg:
# fix(transaction): release connection reference when a server transaction terminates

## Summary

`ServerTx` acquires a connection reference when it is created but never releases
it, so **every incoming request leaks one connection reference** for the lifetime
of the process. `ClientTx` already releases its reference in `delete()`; this PR
adds the symmetric release to `ServerTx` (plus the matching cleanup on the
`Init()` error path).

## Root cause

When the transaction layer creates a server transaction it obtains the
connection through `serverRequestConnection`, which calls the pool's
`GetConnection` — and `GetConnection` **increments the connection's reference
count** (`transport_connection_pool.go`: *"Getting connection pool increases
reference. Make sure you TryClose after finish"*).

The client side balances this:

```go
// transaction_client_tx.go — ClientTx.delete()
if _, err := tx.conn.TryClose(); err != nil {
    tx.log.Info("Closing connection returned error", "error", err, "tx", tx.Key())
}
```

The server side does not. `ServerTx.delete()` stops its timers, closes `done`,
and fires `onTerminate` — but never calls `tx.conn.TryClose()`. And
`serverTxTerminate` only does `serverTransactions.drop(key)`. So the `Ref(+1)`
taken at creation is never matched by a `Ref(-1)`.

Net effect: `ref(connection)` grows by one per received request and never comes
back down. The connection's reference count can therefore never reach zero, so
the pool's reference-based cleanup will **never close that connection**, even
when it should (e.g. when the peer rotates / the registration moves to a new
connection). On long-lived UAs that receive steady in-dialog traffic (NOTIFY
streams, re-INVITEs, OPTIONS keepalives) the count climbs without bound, and
connections that ought to be reaped are pinned open.

## Reproduction / evidence

On a production proxy (a UA holding a persistent TLS registration that receives a
steady NOTIFY stream), counting the debug ref accounting over a 25-minute window:

```
TCP reference increments = 1541
TCP reference decrements =  870

client transactions: 869 created / 870 destroyed
server transactions: 672 created / 672 destroyed

1541 == 869 (client created) + 672 (server created)   # every tx Refs(+1)
 870 ==                         870 (client destroyed)  # only client tx TryCloses
leaked ≈ 671 ≈ 672 == server transactions created
```

The leaked count tracks the number of *server* transactions exactly — i.e. the
number of received requests — and a healthier (busier) peer leaks **faster**, not
slower.

## Fix

Two small changes, both mirroring the existing client-side handling:

1. `ServerTx.delete()` (`sip/transaction_server_tx.go`) — release the connection
   reference, exactly as `ClientTx.delete()` does.
2. `serverTxCreate()` (`sip/transaction_layer.go`) — if `tx.Init()` fails the
   transaction is discarded before it can ever reach `delete()`, so release the
   reference on that error path too (mirrors the `conn.TryClose()` that
   `clientTxRequest` performs on its "transaction already exists" path).

`delete()` is guarded by the once-only `tx.closed` flag, so `TryClose()` is
called at most once per transaction — exactly balancing the single `Ref(+1)`
taken at creation.

## Test

`TestServerTxReleasesConnectionRefOnTerminate` (in
`sip/transaction_server_tx_test.go`) uses a minimal `Connection` that records
`Ref`/`TryClose`. It simulates the reference the layer takes at creation, builds
a `ServerTx`, terminates it, and asserts the reference returns to zero (one
`TryClose`).

Verified against `sip` at this revision:

- Without the fix: `FAIL` — `Terminate must TryClose the connection exactly once:
  expected 1, actual 0`.
- With the fix: `ok` — and the full `go test ./sip/` suite passes.

Note: `ServerTx.delete` nil-checks `tx.conn` before `TryClose`, because some
lower-level tests construct a transaction with a `nil` connection
(`TestServerTransactionContext`). Real transactions always have a non-nil
connection from `serverRequestConnection`.

## Impact / compatibility

- No public API change.
- No behavioural change for connections that are already long-lived and shared:
  `TryClose()` only closes the underlying socket when the ref count reaches zero;
  for a still-referenced connection it just decrements the counter.
- Restores the documented contract of `GetConnection` ("Make sure you TryClose
  after finish") for the server path.

## Checklist

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] Mirrors existing `ClientTx` reference handling
- [x] Added regression test


sip/transaction_server_tx_refleak_test.go
```diff
@@ -0,0 +1,49 @@
+package sip
+
+import (
+	"log/slog"
+	"net"
+	"sync"
+	"testing"
```
emiago: No new file. Also try to reuse existing connections wrappers with package fakes

sip/transaction_server_tx.go
```diff
@@ -262,5 +262,16 @@ func (tx *ServerTx) delete(err error) bool {
 	if onterm != nil {
 		onterm(key, err)
 	}
+
+	// Release the connection reference taken at creation (serverRequestConnection
+	// -> GetConnection increments it). ClientTx.delete does the same; without it
+	// every received request leaks one connection ref and the pool can never reap
+	// the connection. Guarded once by tx.closed above; nil-checked because some
+	// tests construct a tx with a nil conn (TestServerTransactionContext).
+	if tx.conn != nil {
```
emiago, line 268: ok yes make sense. Same as client transcation. No need really for above comment

emiago:
Hi @simonrakcsg . Let your AI check my review.

simonrakcsg:
Thanks for the review — both addressed in the latest push:

- **No new file / reuse package fakes:** folded the regression test into `transaction_server_tx_test.go` as `TestServerTransactionReleasesConnRef`. It now uses a real `UDPConnection` wrapping `fakes.UDPConn` (like the other tests in that file) and asserts on the connection's own refcount via `conn.Ref(0)` instead of a bespoke fake. Verified it fails without the fix (ref stays at 2) and passes with it.
- **Drop the comment:** trimmed the `ServerTx.delete` comment down to a one-liner noting it mirrors `ClientTx.delete` (the nil-check note stays since a couple of tests build a tx with a nil conn).

Full `go test ./sip/` is green.

PR #310, opened 2026-06-04
feat(transport): KeepaliveHandler to observe RFC 5626 CRLF keepalives

simonrakcsg:
# feat(transport): KeepaliveHandler to observe RFC 5626 CRLF keepalives

## Summary

Adds an optional `KeepaliveHandler` callback, set via
`WithTransportLayerKeepaliveHandler(...)`, that fires when a connection-oriented
transport (TCP / TLS / WS / WSS) receives an RFC 5626 §3.5.1 CRLF keepalive. The
callback receives the connection's `TransportReadProps` and an `isPing` flag:

- `isPing == true`  — a double-CRLF `"\r\n\r\n"` ping (which the transport still
  auto-pongs, unchanged).
- `isPing == false` — a single-CRLF `"\r\n"` pong (the peer's reply to *our* ping).

It fires at the transport's *own* keepalive-detection point — the same `if
len(bytes.Trim(data, "\r\n")) == 0` branch that already decides whether to pong —
so an application can observe keepalives authoritatively without re-implementing
CRLF framing.

## Motivation

sipgo already detects and auto-pongs inbound CRLF pings, but it does not expose
*that an inbound keepalive happened*. We need that signal to track **round-trip
liveness** of an idle outbound flow: a client sends a `"\r\n\r\n"` ping
(RFC 5626 client keepalive) and wants to confirm the `"\r\n"` pong came back, as
proof the flow is alive end-to-end — distinct from a local write merely
succeeding, which proves nothing on a half-open / blackholed socket.

The obvious alternative — a `WithTransportLayerReadFilter` that sniffs for CRLF
bytes — is unreliable: the read filter runs on raw, unframed read chunks, so a
ping split across two reads false-positives and a pong coalesced with following
SIP bytes is missed. The keepalive detection inside the transport is the correct,
canonical place to observe this; this PR simply surfaces it. (It is exactly as
reliable as sipgo's existing decision to pong, because it *is* that same branch.)

## Scope

Wired for the connection-oriented transports where "this connection received a
keepalive" is meaningful and the connection props are readily in scope: **TCP**
(and **TLS**, which embeds `TransportTCP`) and **WS** (and **WSS**, which embeds
`TransportWS`). UDP is intentionally excluded — it is connectionless and its
CRLF check sits in a propless `parseAndHandle` helper; a keepalive-received
signal there carries little meaning.

The handler runs on the connection read goroutine; the doc comment notes it must
be fast and non-blocking. It is purely additive: behavior is unchanged when no
handler is set, and the existing auto-pong path is untouched.

## Changes

- `sip/transport.go` — new `KeepaliveHandler func(info TransportReadProps, isPing bool)` type + doc.
- `sip/transport_layer.go` — `keepaliveHandler` field, `WithTransportLayerKeepaliveHandler`
  option, and propagation to TCP/TLS/WS/WSS (mirrors the existing `readFilter` wiring,
  including the `withTransports` set-if-nil path).
- `sip/transport_tcp.go` / `sip/transport_ws.go` — `keepaliveHandler` field + the fire
  site in the CRLF-keepalive branch (TCP fires before its auto-pong write).
- `sip/transport_keepalive_test.go` — drives `readConnection` over a `net.Pipe`,
  asserts the handler fires with `isPing=false` for `"\r\n"` and `isPing=true` for
  `"\r\n\r\n"`, and that a keepalive is never forwarded to the SIP message handler.

## Testing

```
go test ./sip/ -run TestTransportLayerKeepaliveHandlerTCP -race -count=1
go test ./sip/ -count=1   # full package, no regressions
go vet ./sip/
```

All green against current `main`. The patch applies cleanly to a fresh upstream
checkout.

## Compatibility

Additive only — new optional type, field, and option. No existing signature or
behavior changes; `nil` handler is a no-op.


emiago:
Hi @simonrakcsg AI. Can you make it short, what is goal here? Why read filter is not enough?

simonrakcsg:
Hi @emiago — thanks for looking.

**Goal:** confirm an idle outbound flow is alive end-to-end by observing that our `\r\n\r\n` ping got its `\r\n` pong back. A local write succeeding proves nothing on a half-open/blackholed socket, and an OPTIONS ping isn't always available as an alternative.

**Why not the read filter:** it *can* see the bytes, but it's a data-path transformer — it returns bytes and an error that tears the connection down — so using it for passive observation is the wrong semantics, and you'd have to re-implement the CRLF heuristic (and the ping-vs-pong length check) that the transport already does one line later. The handler just surfaces sipgo's own authoritative keepalive branch — including the ping/pong distinction it already computes — as a no-op-when-unset callback.

PR #311, opened 2026-06-12
fix(dialog): ACK to 2xx must not copy Route from INVITE

ddas-rt:
## Summary

- For a 2xx ACK, RFC 3261 §12.1.2 + §12.2.1.1 require the route-set to be derived from reversing the Record-Route headers in the 200 OK response, not copied from the original INVITE. `buildReq()` already handles this correctly when `WriteAck → WriteRequest` is called, so `newAckRequestUAC` must not also copy the INVITE Route headers (which caused duplicates and wrong `Destination()`).
- For a non-2xx ACK (§17.1.1.3), only the INVITE Route headers should be copied; removed the incorrect `else` branch that was also adding reversed Record-Route headers.
- Fixes `TestDialogClientRequestRecordRouteHeaders` (LooseRouting + StrictRouting subtests).

## Test plan

- [x] `go test ./...` passes green
- [x] `TestDialogClientRequestRecordRouteHeaders/LooseRouting` passes
- [x] `TestDialogClientRequestRecordRouteHeaders/StrictRouting` passes

### DIFF
--- a/dialog_client.go
+++ b/dialog_client.go
@@ -499,10 +499,12 @@ func newAckRequestUAC(inviteRequest *sip.Request, inviteResponse *sip.Response, 
 	)
 	ackRequest.SipVersion = inviteRequest.SipVersion
 
-	if len(inviteRequest.GetHeaders("Route")) > 0 {
-		sip.CopyHeaders("Route", inviteRequest, ackRequest)
-	}
-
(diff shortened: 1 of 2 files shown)


ddas-rt:
@emiago Please review. Happy to clarify further if needed.

emiago:
LGTM. Nice fix

ddas-rt:
@emiago any ETA on the next release?

PR #313, opened 2026-07-15
fix: guard connCloser with a mutex to avoid a data race on cancel

csiwek:
Synchronize connCloser access in ListenAndServe

Without this,  -race detector flags causes data races on cancel. 
It will probably never happens when running single ListenAndServe  but it is possible when you have multiple sockets open.

Also needed if you just test code that uses sipgo with -race flag

### DIFF
--- a/server.go
+++ b/server.go
@@ -8,6 +8,7 @@ import (
 	"log/slog"
 	"net"
 	"strings"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


csiwek:
I don't think the failing test is related to the PR

emiago:
@csiwek added better fix. Check main if it works.

PR #316, opened 2026-07-18
Fix data race on listener close in ListenAndServe

### LINKED ISSUE #315: Data race on connCloser in Server.ListenAndServe
**Version:** v1.4.3 (also present on `master` as of 2026-07-18)
**Go:** 1.26.5 darwin/arm64

## Description

`Server.ListenAndServe` declares `connCloser` and immediately spawns a goroutine
that reads it on context cancellation. The variable is only assigned *after*
that goroutine is running, and the two accesses are unsynchronised — no mutex,
atomic, or channel.

```go
func (srv *Server) ListenAndServe(ctx context.Context, network string, addr string) error {
	network = strings.ToLower(network)
	var connCloser io.Closer

	// TODO consider different design to avoid this additional go routines
	go func() {
		select {
		case <-ctx.Done():
			if connCloser == nil {      // <-- read
				return
			}
			if err := connCloser.Close(); err != nil {
				srv.log.Error("Failed to close listener", "error", err)
			}
		}
	}()

	switch network {
	case "udp", "udp4", "udp6":
		...
		udpConn, err := net.ListenUDP(network, laddr)
		if err != nil {
			return fmt.Errorf("listen udp error. err=%w", err)
		}

		connCloser = udpConn            // <-- write
		listenReadyCtx(ctx, network, udpConn.LocalAddr().String())
		return srv.tp.ServeUDP(udpConn)
```

In v1.4.3 that is the read at `server.go:105` / `:108` against the write at
`server.go:125`. The same shape applies to every transport branch (TCP, TLS, WS,
WSS), not just UDP.

## Race detector output

```
WARNING: DATA RACE
Read at 0x00c000188020 by goroutine 11:
  github.com/emiago/sipgo.(*Server).ListenAndServe.func1()
      sipgo@v1.4.3/server.go:105 +0x58

Previous write at 0x00c000188020 by goroutine 10:
  github.com/emiago/sipgo.(*Server).ListenAndServe()
      sipgo@v1.4.3/server.go:128 +0x6c4
```

A second race follows from the same cause — the closer goroutine calling
`Close()` on the `net.Conn` while `net.newUDPConn` is still initialising it:

```
WARNING: DATA RACE
Read at 0x00c0001a2000 by goroutine 11:
  net.(*conn).ok()
  net.(*conn).Close()
  github.com/emiago/sipgo.(*Server).ListenAndServe.func1()
      sipgo@v1.4.3/server.go:108 +0x78

Previous write at 0x00c0001a2000 by goroutine 10:
  net.newUDPConn()
  net.ListenUDP()
  github.com/emiago/sipgo.(*Server).ListenAndServe()
      sipgo@v1.4.3/server.go:123 +0x5e8
```

## Reproduction

Call `ListenAndServe` under `-race` and cancel the context shortly after start:

```go
ctx, cancel := context.WithCancel(context.Background())
go srv.ListenAndServe(ctx, "udp", "127.0.0.1:0")
time.Sleep(10 * time.Millisecond)
cancel()
```

Reproduces reliably for us in a test suite that starts and stops a server per
test case.

## Impact

Beyond being UB, the practical failure is at shutdown:

- If the goroutine observes the pre-write `nil`, it returns early and the
  listener is **never closed** — the socket leaks for the process lifetime. In a
  test suite that starts a server per case, this exhausts ports.
- If it observes a torn interface value, `Close()` is called on a partially
  constructed `net.Conn`.

Because the read only happens on `ctx.Done()`, this is mostly invisible in
long-running servers and shows up in tests and in code that restarts listeners.

## Suggested fix

Create the listener before spawning the closer goroutine, so the value is
captured rather than shared. Roughly:

```go
func (srv *Server) ListenAndServe(ctx context.Context, network, addr string) error {
	network = strings.ToLower(network)

	closeOnDone := func(c io.Closer) {
		go func() {
			<-ctx.Done()
			if err := c.Close(); err != nil {
				srv.log.Error("Failed to close listener", "error", err)
			}
		}()
	}

	switch network {
	case "udp", "udp4", "udp6":
		laddr, err := net.ResolveUDPAddr(network, addr)
		if err != nil {
			return fmt.Errorf("fail to resolve address. err=%w", err)
		}
		udpConn, err := net.ListenUDP(network, laddr)
		if err != nil {
			return fmt.Errorf("listen udp error. err=%w", err)
		}
		closeOnDone(udpConn)
		listenReadyCtx(ctx, network, udpConn.LocalAddr().String())
		return srv.tp.ServeUDP(udpConn)
	...
```

That also removes the `connCloser == nil` check entirely, and addresses the
existing `// TODO consider different design to avoid this additional go
routines`.

I have opened a PR implementing this.

## Workaround for others hitting this

Bind the listener yourself and call `ServeUDP` / `ServeTCP` directly, skipping
`ListenAndServe`:

```go
conn, err := net.ListenPacket("udp", addr)
if err != nil {
	return err
}
defer conn.Close()
go func() { <-ctx.Done(); conn.Close() }()
return srv.ServeUDP(conn)
```

andycol:
Fixes #315.

## Problem

`ListenAndServe` and `ListenAndServeTLS` declare a `connCloser` variable and
spawn a goroutine that reads it on `ctx.Done()`, then assign it only after the
listener has been created. The read and the write are unsynchronised.

```go
var connCloser io.Closer

// TODO consider different design to avoid this additional go routines
go func() {
	select {
	case <-ctx.Done():
		if connCloser == nil {   // <-- read
			return
		}
		if err := connCloser.Close(); err != nil {
			...
		}
	}
}()

// ...
connCloser = udpConn             // <-- write
```

Beyond being a data race, this leaks sockets. When the goroutine observes the
pre-assignment `nil` it returns early and **nothing ever closes the listener**.
That is hit most reliably when the context is already cancelled by the time the
listener is created, and it shows up in test suites that start a server per
case, where it eventually exhausts ports.

Because the read only happens on `ctx.Done()`, it stays invisible in
long-running servers and surfaces in tests and in code that restarts listeners.

## Fix

Pass the listener to a small `closeOnDone` helper rather than sharing it through
a variable assigned later. The value is captured, so there is nothing to race
on, and a context that is already done still closes the listener instead of
leaking it.

```go
func (srv *Server) closeOnDone(ctx context.Context, c io.Closer) {
	go func() {
		<-ctx.Done()
		if err := c.Close(); err != nil {
			srv.log.Error("Failed to close listener", "error", err)
		}
	}()
}
```

This also resolves the existing `// TODO consider different design to avoid this
additional go routines` and drops the now-unnecessary `nil` check.

Applied to every transport branch: UDP, TCP, WS, TLS and WSS.

## Tests

Two regression tests in `server_listener_race_test.go`:

- `TestListenAndServeClosesListenerOnCancel` — cancelling the context must close
  the listener.
- `TestListenAndServeCancelledBeforeBind` — the case the old code leaked most
  reliably.

Both use the existing `ListenReadyCtxKey` / `ListenReadyFuncCtxValue` hook with
port `0`, so they neither hard-code a port nor race the server to bind one.

On current `main` under `-race`:

```
WARNING: DATA RACE
--- FAIL: TestListenAndServeCancelledBeforeBind (3.00s)
```

With this change, passing with `-count=5`:

```
ok  	github.com/emiago/sipgo	1.470s
```

## Notes

Two tests fail on `main` both before and after this change, unrelated to it:
`TestClientWriteRequestRejectsCRLFAfterBuild` and
`TestClientTransactionRequestRejectsCRLFAfterBuild` in the root package, plus
`TestServerTransactionRespondRejectsCRLF` in `sip/`. I have left those alone.

Behaviour is otherwise unchanged: the non-TLS path still closes only on parent
context cancellation, and the TLS path keeps its `defer cancel()` so the
listener is closed when the function returns.

### DIFF
--- a/server.go
+++ b/server.go
@@ -92,25 +92,25 @@ func newBaseServer(ua *UserAgent, options ...ServerOption) (*Server, error) {
 	return s, nil
 }
 
+// closeOnDone closes c once ctx is done.
+//
+// The listener is passed in rather than shared through a variable the caller
+// assigns later, so there is no race between this goroutine's read and that
+// assignment. It also means a context already cancelled by the time the
+// listener exists still closes it, instead of leaking the socket.
+func (srv *Server) closeOnDone(ctx context.Context, c io.Closer) {
+	go func() {
+		<-ctx.Done()
+		if err := c.Close(); err != nil {
+			srv.log.Error("Failed to close listener", "error", err)
+		}
+	}()
+}
+
 // Serve will fire all listeners
 // Network supported: udp, tcp, ws
 func (srv *Server) ListenAndServe(ctx context.Context, network string, addr string) error {
 	network = strings.ToLower(network)
-	var connCloser io.Closer
-
-	// TODO consider different design to avoid this additional go routines
-	go func() {
-		select {
-		case <-ctx.Done():
-			if connCloser == nil {
-				return
-			}
-			if err := connCloser.Close(); err != nil {
-				srv.log.Error("Failed to close listener", "error", err)
-			}
-
-		}
-	}()
 
 	switch network {
 	case "udp", "udp4", "udp6":
@@ -125,7 +125,7 @@ func (srv *Server) ListenAndServe(ctx context.Context, network string, addr stri
 			return fmt.Errorf("listen udp error. err=%w", err)
 		}
 
-		connCloser = udpConn
+		srv.closeOnDone(ctx, udpConn)
 		listenReadyCtx(ctx, network, udpConn.LocalAddr().String())
 		return srv.tp.ServeUDP(udpConn)
 
@@ -140,7 +140,7 @@ func (srv *Server) ListenAndServe(ctx context.Context, network string, addr stri
 			return fmt.Errorf("listen tcp error. err=%w", err)
 		}
 
-		connCloser = conn
+		srv.closeOnDone(ctx, conn)
 		listenReadyCtx(ctx, network, conn.Addr().String())
 
 		return srv.tp.ServeTCP(conn)
@@ -157,7 +157,7 @@ func (srv *Server) ListenAndServe(ctx context.Context, network string, addr stri
 			return fmt.Errorf("listen tcp error. err=%w", err)
 		}
 
-		connCloser = conn
+		srv.closeOnDone(ctx, conn)
 		listenReadyCtx(ctx, network, conn.Addr().String())
 		// and uses listener to buffer
 		return srv.tp.ServeWS(conn)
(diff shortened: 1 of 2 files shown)


emiago:
it is already pointet out with #313 and fixed very similar on main branch

PR #330, opened 2026-07-19
dialog: answer a stale nonce re-challenge on INVITE

### LINKED ISSUE #329: sip: stale=true re-challenge on INVITE is rejected instead of answered
WaitAnswer guards the digest retry on the absence of an Authorization header (dialog_client.go:307). After the first attempt that header is present, so a second 401 is never treated as a challenge and the call fails.

That guard is also doing loop-termination duty, which is why it cannot answer a re-challenge — one guard, two jobs.

It matters for stale=true. RFC 2617 3.2.1 defines stale as "the nonce is old, retry with the new one and the same credentials", and RFC 3261 22.4 adopts the digest scheme for SIP. A registrar or proxy rotating nonces sends exactly that, and the INVITE fails on a correct password.

Separately, any 401/407 is treated as a challenge even when it carries no WWW-Authenticate or Proxy-Authenticate header, so a plain rejection surfaces as "No WWW-Authenticate header present" rather than its real status.

Fix: count attempts instead — cap 2, shared by the 401 and 407 arms so a 407 followed by a 401 consumes both — and treat a 401/407 as a challenge only when the matching authenticate header is present. Replacing credentials is safe, since digestAuthApply and digestProxyAuthApply remove the existing authorization header before appending. Exhausting the cap returns the final response as the ErrDialogResponse it returns today, so no new API.

One known weakness: the counter is blunt. It does not read digest.Challenge.Stale, so a genuinely wrong password costs two round trips instead of one, which on some PBXes feeds lockout counters. icholy/digest exposes Challenge.Stale at challenge.go:18 if you would rather gate on it.

Mliviu79:
Fixes #329

`WaitAnswer` guarded the digest retry on the absence of an `Authorization` header, so after the first attempt a second 401 was never treated as a challenge. That guard was also doing loop-termination duty, which is why it could not answer a re-challenge.

A `stale=true` re-challenge (RFC 2617 3.2.1, adopted for SIP by RFC 3261 22.4) asks the UAC to retry with the new nonce and the same credentials, so a registrar rotating nonces failed the INVITE on a correct password.

Attempts are counted instead, capped at two and shared by the 401 and 407 arms so a 407 followed by a 401 consumes both. Replacing credentials is safe because `digestAuthApply` and `digestProxyAuthApply` remove the existing authorization header before appending.

A 401/407 is treated as a challenge only when it carries the matching authenticate header, so a plain rejection reaches the caller as its own status rather than "No WWW-Authenticate header present".

**No new exported API** — `maxAuthAttempts` is unexported and exhausting the cap returns the same `ErrDialogResponse` as today, so terminal behaviour is unchanged for existing callers. Your `AuthLoop` subtest passes unmodified; three subtests are added alongside it for the stale re-challenge, cap exhaustion, and the no-challenge case.

### DIFF
--- a/dialog_client.go
+++ b/dialog_client.go
@@ -228,6 +228,12 @@ var (
 	WaitAnswerForceCancelErr = errors.New("Context cancel forced")
 )
 
+// maxAuthAttempts caps how many digest challenges are answered on one INVITE.
+// Two allows a stale=true re-challenge (RFC 2617 3.2.1, adopted for SIP by RFC
+// 3261 22.4) to be answered with the
+// fresh nonce, while still terminating an endless challenge stream.
+const maxAuthAttempts = 2
+
 // WaitAnswer waits for success response or returns ErrDialogResponse in case non 2xx
 // Canceling context while waiting 2xx will send Cancel request. It will block until 1xx provisional is not received
 // If Canceling succesfull context.Canceled error is returned
@@ -238,6 +244,9 @@ func (s *DialogClientSession) WaitAnswer(ctx context.Context, opts AnswerOptions
 	tx, inviteRequest := s.inviteTx, s.InviteRequest
 	var r *sip.Response
 	var err error
+	// Counts challenges answered on this INVITE, shared by both arms so that a
+	// 407 followed by a 401 consumes both attempts.
+	authAttempts := 0
 	for i := 0; ; i++ {
 		if i > 10 {
 			// Preventing some long loops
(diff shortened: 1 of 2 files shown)


emiago:
I have run analysis and found some things below. (analysis may be wrong)

  - RemoveHeader removes only the first exact-case header, so multiple credentials can remain/stack.
  - Replacing all proxy credentials would be wrong for multiple proxy realms; SIP requires credentials for other
    realms to remain usable. RFC 3261 §22.3 (https://datatracker.ietf.org/doc/html/rfc3261.html#section-22.3)

  - The retry cap allows a second attempt for any challenge, including stale=false/wrong-password responses. It
    should likely retry again only when the parsed challenge says stale=true.

  - Tests cover repeated 401, but not repeated 407 or preservation of multiple Proxy-Authorization headers.


 so I bit simplified and re implement a patch.  Pls recheck.
 Thanks for using sipgo.

PR #332, opened 2026-07-19
transport: pool a WSS connection under the remote address

### LINKED ISSUE #331: sip: WSS pools the connection under the local address, not the remote
TransportWSS.CreateConnection calls the pool with (laddr, raddr):

    sip/transport_wss.go:53   t.pool.addSingleflight(laddr, raddr, ...)

UDP, TCP, TLS and WS all pass (raddr, laddr). The pool keys on its first argument, so a WSS socket is registered under our own local address — twice, since the pool also keys c.LocalAddr() — and under the far end address never.

The lookup that matters is serverRequestConnection -> GetConnection(req.Source()) for an inbound request. It misses a connection that is open and healthy. A WebSocket client is not reachable at its contact address, so that socket is the only path back to it and the INVITE is undeliverable.

Fix: pass raddr first, matching the other four transports.

Mliviu79:
Fixes #331

`TransportWSS.CreateConnection` passed `(laddr, raddr)` to the pool while UDP, TCP, TLS and WS all pass `(raddr, laddr)`. The pool keys on its first argument, so the socket was registered under our own local address — twice, since the pool also keys `c.LocalAddr()` — and under the far end address never.

`serverRequestConnection` looks a connection up by `req.Source()` for an inbound request, so it missed one that was open and healthy. A WebSocket client is not reachable at its contact address, so that socket is the only path back to it and the request was undeliverable.

One line of source. The test drives `CreateConnection` against a real TLS WebSocket server on loopback with certificate verification left on, and fails on the old order — the existing pool unit tests all pass the arguments in by hand, so none of them can catch a caller composing correct pieces in the wrong order.

### DIFF
--- a/sip/transport_wss.go
+++ b/sip/transport_wss.go
@@ -50,7 +50,12 @@ func (t *TransportWSS) CreateConnection(ctx context.Context, laddr Addr, raddr A
 		return nil, fmt.Errorf("remote address IP not resolved")
 	}
 
-	conn, err := t.pool.addSingleflight(laddr, raddr, t.connectionReuse, func() (Connection, error) {
+	// raddr first. The pool keys on its first argument, so passing laddr there
+	// registered this socket under our own local address -- twice, since the pool
+	// also keys the local address -- and under the far end's address never,
+	// leaving every lookup driven by the peer with no connection to find. UDP,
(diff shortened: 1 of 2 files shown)


emiago:
fixed simpler

PR #333, opened 2026-07-19
transport: configurable write and read timeouts on TCP transport

Mliviu79:
Picks up #268, in the shape you asked for there:

> Right now you can customize transports trhough `TransportsConfig` which can be passed through `UA` on TransportLayer. What we basically need is to add those options somowhere in tcp transport.

`TransportTCP` gains `WriteTimeout` and `ReadTimeout`, so both are reachable through `TransportsConfig` with no package globals. `TransportTLS` embeds `*TransportTCP`, so its dial path copies `WriteTimeout` into the connection it builds and it picks up `ReadTimeout` by reusing `readConnection`.

Zero is the default and arms no deadline on either path, so this is a no-op for every existing user — which is the answer to "I find this very hardcoded": nothing is hardcoded and nothing changes unless a caller sets it.

**Write side.** A write to a peer that stops draining its receive window blocks the sending goroutine indefinitely. When `WriteTimeout` is positive, each write arms `SetWriteDeadline`.

**Read side.** A peer that opens a connection and sends nothing, or stalls midway through a message, holds a goroutine, a socket and a read buffer for as long as it likes. `ParseMaxMessageLength` bounds how much memory one message costs but nothing bounded a peer in time. The read loop arms `SetReadDeadline` per iteration and returns on a `net.Error` reporting `Timeout()`, after which the existing deferred `pool.CloseAndDelete` tears the connection down. RFC 5626 keep-alives reset the deadline, so a healthy long-lived connection never reaches it.

On the "could TCP keep alive solve it" question from #268 — it does not cover the write side at all, and on the read side it only detects a *dead* peer, not a live one that is deliberately slow or silent. That is the case these bound.

Credit to @j-licht for #268, whose diff this supersedes. Happy to close this if you would rather they carry it.

Tests cover both timeouts, propagation from transport to connection, an idle peer releasing the read goroutine over a real `net.Pipe`, and a stalled peer on the write side.

### DIFF
--- a/fakes/tcp_conn.go
+++ b/fakes/tcp_conn.go
@@ -19,6 +19,7 @@ type TCPConn struct {
 	mu sync.Mutex
 
 	writeDeadlines []time.Time
+	readDeadlines  []time.Time
 }
 
 // SetWriteDeadline records the deadline instead of enforcing it. The embedded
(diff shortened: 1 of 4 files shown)


emiago:
Fixed as 
https://github.com/emiago/sipgo/pull/268. 
I suggest opening issue if this does not satisfy/fixes problem.

PR #335, opened 2026-07-19
sip: add Session-Expires and Min-SE headers

### LINKED ISSUE #334: sip: no typed access to Session-Expires and Min-SE
RFC 4028 session timers need Session-Expires and Min-SE. Neither has a typed header in the sip package, so a caller reads them through GetHeader and parses the delta and the refresher param itself.

Proposal: SessionExpiresHeader (uint32 delta plus optional HeaderParams) and MinSEHeader (uint32 delta), both implementing Header, reachable through lazy accessors over the generic header passthrough — the same shape as ReferTo and ReferredBy. The Session-Expires accessor also matches the compact form "x".

Params such as refresher=uac|uas are stored as sent rather than interpreted, and a malformed value makes the accessor return nil. Electing a refresher and enforcing the Min-SE floor are negotiation policy and stay with the caller.

headersParsers is unchanged — parsing happens on demand, so a message without session timers does no extra work.

This is additive API, so it is a design call rather than a defect.

Mliviu79:
Fixes #334

Adds `SessionExpiresHeader` (uint32 delta-seconds plus optional `HeaderParams`) and `MinSEHeader` (uint32 delta). Both implement `Header` — `Name`, `Value`, `String`, `StringWrite`, `headerClone` — and `SessionExpiresHeader` also gets a typed `Clone`.

### DIFF
--- a/sip/headers.go
+++ b/sip/headers.go
@@ -443,6 +443,27 @@ func (hs *headers) ReferredBy() *ReferredByHeader {
 	return nil
 }
 
+// SessionExpires parses underlying RFC 4028 Session-Expires header or nil if not
+// exists. Nil is also returned when the value is malformed, which is the read
+// side of a peer that offers no session timers: the dialog then runs with no
+// refresh loop instead of failing.
+func (hs *headers) SessionExpires() *SessionExpiresHeader {
+	h := &SessionExpiresHeader{}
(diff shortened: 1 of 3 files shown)


emiago:
Move the test sto existing files, and reduce them to minimal. Just verifying that parsing and writing is enough. Header has no bigi impact on rest of lib

PR #341, opened 2026-07-19
transport: log parse failures at Debug without the payload

### LINKED ISSUE #340: sip: parse failures log the raw payload at Error, pre-auth
TCP, UDP and WS all log a failed parse as:

    t.log.Error("failed to parse", "error", err, "data", string(data))

Both the level and the payload are reachable by any host that can reach the listener, before anything has authenticated.

On the level: unparseable input is the normal background of an open SIP port, and the peer picks when it fires. At Error a remote host sets the error rate of the process.

On the payload: a REGISTER that fails to parse still carries its Authorization header, so the digest response reaches the log (RFC 3261 20.7). A message that fails on its CSeq — the shape of a REGISTER retry after a 401 — logs the credential in full.

Fix: log at Debug with src and a bytes count instead of the raw payload. That is still enough to chase a genuinely malformed peer. TLS and WSS have no parse call site of their own and inherit this through the TransportTCP and TransportWS embeds.

Mliviu79:
Fixes #340

### DIFF
--- a/sip/transport_parselog_test.go
+++ b/sip/transport_parselog_test.go
@@ -1,0 +1,131 @@
+package sip
+
+import (
+	"context"
+	"fmt"
+	"log/slog"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// logCapture is a slog.Handler retaining every record so a test can assert on
+// the level and the attributes a code path emits. Enabled reports true for all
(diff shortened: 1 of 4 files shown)


emiago:
I had recent fixes (AI analysis) check is conflict because of this. Again remove comments, we do not need to explain debug lines

emiago:
I see no point having tests on log ouput.

PR #343, opened 2026-07-19
parser: bound datagram body allocation by message size

### LINKED ISSUE #342: sip: Parser.Parse allocates the declared Content-Length before rejecting the message
On the datagram path `Parser.Parse(data, false)` takes the declared `Content-Length` into `make([]byte, bodySize)` guarded only by `if bodySize < 0`. The entry check bounds the bytes that arrived, not the body that was declared.

Measured on main (3f46033), default `ParseMaxMessageLength` = 65,535, a body-less INVITE declaring `Content-Length: 4294967295`:

    datagram on the wire   229 bytes
    heap allocated         4,294,974,824 bytes
    body attached          4,294,967,295 bytes
    returned error         reading body incomplete

The message is rejected, but only after the allocation. UDP and WebSocket both reach this through `ParseSIP` -> `Parse(data, false)`.

Fix: bound the declared body against what is left of the message budget, before the `make`.

Mliviu79:
Fixes #342

One check, between the `bodySize == 0` early return and the `make`:

```go
if bodySize > p.MaxMessageLength-total {
    return msg, total, ErrMessageTooLarge
}
```

### DIFF
--- a/sip/parser.go
+++ b/sip/parser.go
@@ -275,6 +275,15 @@ func (p *Parser) Parse(data []byte, stream bool) (Message, int, error) {
 	if bodySize == 0 {
 		return msg, total, nil
 	}
+	// Content-Length is attacker-controlled and independent of how many bytes
+	// actually arrived: a small datagram can declare a body of up to 4 GB and
+	// drive the make below to allocate it before the copy can reject it. The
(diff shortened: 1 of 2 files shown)


emiago:
pls remove comments. It is clear what is issue here. We solved this on streaming, not sure why I missed this.

PR #345, opened 2026-07-19
transport: close TCP connection on any stream framing error

### LINKED ISSUE #344: sip: the TCP read loop continues after a framing error and grafts the next request onto the failed one
`readConnection` ignores `parseStream`'s outcome and reads on, and `ParserStream.ParseNext` sets `reset := err == nil`, so a message that failed to parse keeps its half-built state. The next bytes are parsed as its headers.

Measured on main (3f46033), two requests driven through `readConnection` over a `net.Pipe` — the first with no `Content-Length`, which the stream parser cannot frame, the second a well-formed REGISTER:

    delivered to the handler: 1 message
      Method=INVITE  R-URI=sip:victim@target
      Via     = SIP/2.0/TCP evil;branch=z9hG4bKFIRST
      From    = <sip:alice@target>;tag=2
      Call-ID = SECOND-legit@target
      CSeq    = 2 REGISTER

The two arrive as one request that neither peer sent: the first peer's method and R-URI carrying the second's Call-ID, From-tag and CSeq. The REGISTER is never delivered as itself.

`AppendHeader` keeps the first Via but overwrites From/To/Call-ID/CSeq, which is why the delivered message has a split identity.

Fix: close the connection on any parse error. On a stream transport `Content-Length` is the framing, so a parse error means there is no boundary left to resync on.

Mliviu79:
Fixes #344

Stacked on #324, which makes `parseStream` return an error at all — on main it returns nothing, so `readConnection` cannot react to a parse failure. Please take that first — the diff here is only the widened condition plus a test on top.

### DIFF
--- a/sip/parser_stream.go
+++ b/sip/parser_stream.go
@@ -122,6 +122,12 @@ func (p *ParserStream) Write(data []byte) (int, error) {
 
 // ParseNext parses the next SIP message from an internal buffer.
 // It may return io.ErrUnexpectedEOF, indicating that more data needs to be written with Write.
+//
+// ErrMessageTooLarge is returned in two shapes. If the oversized message was
+// parsed in full, it is fully consumed from the buffer and the stream can be
+// recovered by simply calling ParseNext again. If it is still incomplete, the
+// stream cannot be recovered: there is no message boundary to skip to, so the
+// buffer is drained and the caller should close the connection.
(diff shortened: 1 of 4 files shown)


emiago:
fix conflicts. but also pls tell your AI to make this comments in one sentence. Code should it self be clear enough to understand what is doing and not comments

emiago:
Fixed with other PRs

PR #346, opened 2026-07-28
BYE: Don't copy Route header(s) from the initial INVITE

ddas-rt:
The route set of a BYE request is constructed in the following way:
- For the UAS side, the dialog route set is the list of URIs in the Record-Route header field from the request, taken in order. If no Record-Route exists, the route set is empty. It also says this route set overrides any pre-existing route set.
- For the UAC side, the dialog route set is the list of URIs in the Record-Route header field from the response, taken in reverse order. If no Record-Route exists, the route set is empty. This also overrides any pre-existing route set.

1. So if the CALLEE originally received:

```INVITE sip:bob@example.com SIP/2.0
Record-Route: <sip:p2.example.com;lr>
Record-Route: <sip:p1.example.com;lr>
Contact: <sip:alice@uac.example.com:5060>
```
The callee’s dialog state becomes(SAME ORDER):

`remote target = sip:alice@uac.example.com:5060`
`route set     = <sip:p2.example.com;lr>, <sip:p1.example.com;lr>`

When the callee sends BYE, it uses that dialog state:

```
BYE sip:alice@uac.example.com:5060 SIP/2.0
Route: <sip:p2.example.com;lr>,
Route: <sip:p1.example.com;lr>
```

2. And if the CALLER receives:

```
SIP/2.0 200 OK
Record-Route: <sip:p2.example.com;lr>
Record-Route: <sip:p1.example.com;lr>
Contact: <sip:bob@uas.example.com:5060>
```

The caller’s dialog state becomes(REVERSED):

`remote target = sip:bob@uas.example.com:5060`
`route set     = <sip:p1.example.com;lr>, <sip:p2.example.com;lr>`

When the caller sends BYE, it builds the Request-URI and Route headers from that dialog state. Since the first route has ;lr, this is loose routing, so the BYE looks like:

```
BYE sip:bob@uas.example.com:5060 SIP/2.0
Route: <sip:p1.example.com;lr>
Route: <sip:p2.example.com;lr>,
```


ddas-rt:
@emiago Please review. Missed this BYE change in the previous ACK related [pr](https://github.com/emiago/sipgo/pull/311).

ddas-rt:
There's a test failure in `TestServerTransactionRespondRejectsCRLF` in `sip/transaction_server_tx_test.go:122`, unrelated my change.

emiago:
Merging! Thanks.

ddas-rt:
@emiago Can we please cut a release?

PR #350, opened 2026-08-21
feat(transport): trace complete SIP messages with real transport name

InnovativeUS:
# feat(transport): trace complete SIP messages

Added SIPMessageTracer, set with SIPMessageTrace, called once per parsed message. On write it is called after the message is fully written. A new ParserStream method ParseSIPStreamRaw has been added and ParseSIPStream now wraps it. SIPDebug and SIPDebugTracer behave as before.

Benchmarked with benchstat: no extra allocations, about 2% more time per read.


emiago:
Pls tell me you read existing code? I can not understand need for this? Open issue if something is missing

PR #353, opened 2026-09-03
sip: add TransportWriteFilter (observe-only write mirror of TransportReadFilter)

daldworth:
## Summary

Adds `sip.TransportWriteFilter`, an observe-only hook that mirrors the existing `TransportReadFilter` on the write side:

```go
type TransportWriteProps struct {
	Transport  string
	LocalAddr  net.Addr
	RemoteAddr net.Addr
}
type TransportWriteFilter func(info TransportWriteProps, data []byte)
func WithTransportLayerWriteFilter(f TransportWriteFilter) TransportLayerOption
```

It is invoked from `UDPConnection.WriteMsg`, `TCPConnection.WriteMsg` (TLS shares it) and `WSConnection.WriteMsg` **after a successful write**, with the exact bytes that went on the wire, and is plumbed through `TransportsConfig` / `withTransports` exactly like `readFilter`. It cannot alter or block the write.

## Motivation

We run a Go B2BUA on sipgo and mirror SIP to Homer (HEP). The read filter already gives us every inbound message before parsing, but there was no symmetric way to see what the transaction layer itself sends — retransmissions, the automatic `100 Trying`, CANCEL/ACK/487 — without wrapping every `net.Conn`/`PacketConn` and re-framing TCP. A write filter at `WriteMsg` sees all of it with the real local/remote addresses.

## Notes

- Doc comments on both filters now say `data` is a pooled buffer recycled when `WriteMsg` returns, so a filter that retains it must copy.
- Tests: `TestTransportLayerWriteFilterUDP` (fires once, exact bytes, addresses), `TestTransportLayerWriteFilterTCPSuccess` (over `net.Pipe`), `TestTransportLayerWriteFilterTCPNotCalledOnFailedWrite`.
- No behavior change when no write filter is registered.


emiago:
Hi @daldworth . Is this AI generated?

daldworth:
@emiago Yes it was. Both the PR and diff were done with Claude (code). I see there are some conflicts. Do you want us to look at those?

emiago:
What Claude thinks about Sip Tracer :)

PR #354, opened 2026-09-05
fix(transaction): keep client responses in order of arrival

alexvrf:
### Problem

A UAC loses provisional responses whenever the peer answers immediately.
With a UAS that sends `180 Ringing` and `200 OK` back to back, the `180`
never reaches the TU in 16 to 20 runs out of 20 — `WaitAnswer`'s
`OnResponse` is called only for the `200`.

In a call this is the lost ringing indication and, when the provisional
carries SDP, the lost early media announcement: the caller hears silence
where the network should be heard. It shows up with auto answering
endpoints, intercoms and gateways that emit `100` and `18x` in one burst.

### Root cause

`TransactionLayer.handleMessage` hands every message to its own goroutine:

    case *Response:
        go txl.handleResponseBackground(msg)

The order of arrival is therefore lost right there. Both goroutines then
race for `fsmMu` in `spinFsmWithResponse`. When the `200` wins, the client
INVITE transaction moves to `inviteStateAccepted`; the `180` that was
received first arrives afterwards, hits the `default` branch of that state
and is silently dropped as a stray response (RFC 6026 7.2 — correct
behaviour for a response that really is stray, but this one is not).

The shared `tx.fsmResp` field makes it worse: the late `180` overwrites it
while the transaction is already in `Accepted`. The `TODO v2` at the top of
`transaction_client_tx_fsm.go` — passing the response through the FSM as
context instead of a shared field — points at the same design issue.

### Fix

Match the response to its transaction in the receiving goroutine, where the
order is still well defined, and let the transaction hand out a ticket that
keeps per-transaction delivery in that order. Spinning the FSM stays in a
separate goroutine, because `fsmPassUp` blocks until the TU consumes the
response and blocking the transport receive loop on that would stall the
connection — which is what the comment in `handleMessage` warns about.

`handleResponse`/`handleResponseBackground` are folded into the new
`handleResponseOrdered`; no exported API changes.

### Test

`TestTransactionLayerResponseOrder` hands two responses to `handleMessage`
one after another, exactly as the transport layer does, and requires the
TU to see `180` before `200`. It fails on current `master` and passes with
this change. It repeats 20 times, because a single iteration would be a
coin flip rather than a deterministic failure.

### Verification

- `go test ./...` passes (`testdata/generate_certs.sh` is needed for the
  root package tests).
- With `TEST_INTEGRATION=1 -race`, `TestTransactionLayerClientTx` and
  `TestTransportLayerClientConnectionReuse` fail with data races both with
  and without this change — pre-existing, not touched here.
- Checked against a real B2BUA test suite (two UAs in one process, media
  and routing stubbed): the provisional loss goes from 16–20 out of 20 to
  0 out of 20, and the suite stays green under `-race`.

### DIFF
--- a/sip/transaction_client_tx.go
+++ b/sip/transaction_client_tx.go
@@ -8,7 +8,11 @@ import (
 
 type ClientTx struct {
 	baseTx
-	responses    chan *Response
+	responses chan *Response
+	// receiveOrder is the tail of the chain of tickets handed out by
+	// receiveTicket. It keeps responses of this transaction reaching the
+	// FSM in the order they were received from the transport.
+	receiveOrder chan struct{}
 	timer_a_time time.Duration // Current duration of timer A.
 	timer_a      *time.Timer
 	timer_b      *time.Timer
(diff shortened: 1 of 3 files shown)


alexvrf:
Hello, Please check my PR

emiago:
@alexvrf it is known behavior. This fix breaks scalability, and I hope you realize there is more to it. 
Anyway FSM 2xx should proceed without issues and I can your AI elaborate more why ealry media with auto answering should cause issue with answering.

Issue #125, opened 2024-08-22
Extendable URI parser


IvanRibakov:
Current URI parsing logic is optimized for performance AND for dealing with SIP/SIPS URI schemes. Unfortunately, this is a limiting factor for SIP services who's business logic needs to deal with other types of URIs: `http`/`https`, `tel`, `urn`, `cid`, etc.

Of course, separate implementation-side parsers may be created by `sipgo` users and SIP message parts that contain URIs (generally speaking those are various headers) can be re-parsed based on individual needs. However, such approach comes at a cost of a more complicated implementation-side code. For example, consider [`Geolocation` header](https://datatracker.ietf.org/doc/html/rfc6442#section-4.1). According to spec it can hold a multitude of different scheme URIs:
- `sip`/`sips`
- `pres`
- `http`/`https`
- `cid`
- etc

It is common for all Geolocation header values to be processed together by some location-based piece of logic. For developer it would be extremely convenient to be able to:
1. have `Geolocation` parsed using some generic address-type header parser similar to From/To header parser (headers that contain one or more address values + parameters)
2. retrieve all Geolocation header values from a SIP message
3. pass the collection of header values to another function for processing

If we assume a custom implementation-side parsing logic (for Geolocation header and URIs contained as values), then it may not seem so different/complicated at first. But once we add other complicating factors like the need to access same header values in several different places in the code then we either need to start caching the custom parsing results or face the penalty of re-parsing non-core headers/values each time we need to access them.

### Proposal

1. Make URI logic extendable by allowing to register custom parsers for different URI schemes (similarly to how header parsing can be extended)
    a. By default ship `sipgo` with parsers for `sip`/`sips` scheme as is the case now
3. Add "unparsed" state to the current URI struct to allow capturing original URI string value without the risk of making bad parsing decisions by the generic/default parser
4. Define a convention for representing other URI types and converting to/from generic `Uri` type
5. Add missing URI fields to the current URI struct. As far as I am aware, only `path` is missing at the moment (consider `http:example.com/foo/bar` URI)

Caused by https://github.com/emiago/sipgo/pull/114 (@emiago thank you for your support on it!)

PR #356, opened 2026-09-06
sip: parse a URN instead of reading its NSS as a port

fridokus:
### The bug

`ParseUri` splits everything after the scheme into user, host and port, so a URN's namespace specific string is read as a port:

```go
u := &sip.Uri{}
err := sip.ParseUri("urn:service:sos", u)
// err          strconv.Atoi: parsing "sos": invalid syntax
// u.Host       "service"
// u.String()   "urn:service"
```

It does not just error — the NSS is dropped and the URI left corrupt. Through `parseRequestLine` the error rejects the whole message, so `INVITE urn:service:sos SIP/2.0` never reaches a handler.

That is the [RFC 5031](https://datatracker.ietf.org/doc/html/rfc5031) emergency call, which [3GPP TS 24.229](https://www.3gpp.org/DynaReport/24229.htm) §5.1.6.8.1 requires the network to accept; `urn:uuid:` (`+sip.instance`) fails identically. Both are `absoluteURI`s, admitted as Request-URIs by [RFC 3261](https://datatracker.ietf.org/doc/html/rfc3261#section-7.1) §7.1 — and `Uri`'s doc comment already says the scheme *"can be tel, urn"*.

### The change

The NSS is opaque by definition ([RFC 2141](https://datatracker.ietf.org/doc/html/rfc2141)), so it goes verbatim into a new `Opaque` field and `StringWrite` renders it back unsplit. `Opaque` stays empty for `sip:`, `sips:` and `tel:`, so nothing else is affected.

It runs to the end of the URI on purpose: `;` is a legal NSS character and `uri-parameters` are SIP/SIPS-only (RFC 3261 §19.1.1), so splitting there would corrupt a valid URN.

A deliberately narrow form of the "unparsed" state proposed in point 3 of #125 — limited to the one scheme whose absence loses a call, and no obstacle to the extendable parser that issue asks for.

### Tests

`TestParseUriURN` (8 subtests) and `TestParseRequestURN`, in the existing style. Mutation-verified: disabling the new branch turns every parse-dependent subtest red and leaves the `sip:` control green. Full suite passes.

### DIFF
--- a/sip/parse_uri.go
+++ b/sip/parse_uri.go
@@ -42,6 +42,21 @@ func uriStateScheme(uri *Uri, s string) (uriFSM, string, error) {
 	for i, c := range s {
 		if c == ':' {
 			uri.Scheme = ASCIIToLower(s[:i])
+
+			// A URN's namespace specific string is opaque (RFC 2141): it has
+			// no user, host or port, and the colons inside it separate
+			// namespaces rather than introducing a port. Handing
+			// "service:sos" on to the user and host states reads "sos" as a
+			// port and fails the whole URI, so a URN is taken verbatim and
+			// the FSM stops here.
+			if uri.Scheme == "urn" {
+				uri.Opaque = s[i+1:]
+				if uri.Opaque == "" {
+					return nil, "", fmt.Errorf("urn with no namespace specific string")
+				}
+				return nil, "", nil
+			}
+
 			return uriStateSlashes, s[i+1:], nil
 		}
 		// Check is c still ASCII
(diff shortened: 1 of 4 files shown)


fridokus:
Superseded by edb4fad — closing.

Your version is the better shape: keeping the scheme-specific part raw for any non-SIP scheme rather than special-casing `urn:`, and private with an accessor so a `Uri` cannot hold both raw and parsed fields. Checked `urn:service:sos` and `urn:uuid:` round trip on main.

Thanks!

PR #357, opened 2026-09-07
sip: only treat a CRLF-only read as a keep-alive between messages

fridokus:
### The bug

`readConnection` treats any read of ≤4 bytes containing only CR and LF as an [RFC 5626](https://datatracker.ietf.org/doc/html/rfc5626#section-3.5.1) §3.5.1 keep-alive, wherever it lands. But a read is whatever TCP chose to deliver, so the CRLF ending a header line can arrive on its own — and is then swallowed, leaving the parser holding a bare CR nothing can complete.

The next message fails to frame with `line has no CRLF` even though it is perfectly well formed, and because the buffer still holds the bytes it choked on, so does every message after it. The connection is deaf for good.

Splitting a real 1568-byte REGISTER at any of its 22 CRLFs reproduces it every time. Seen in production as a proxy that stopped processing SIP entirely, sockets draining normally and CPU idle.

### The change

A keep-alive only counts between messages. `ParseSIPStream` already knows when it is part way through one — it returns `ErrParseSipPartial` until it has drained its buffer — so `parseStream` passes that error up instead of folding it into `nil`, and the read loop keeps the last parse result and checks it.

Guarding on "the parser's buffer is empty" instead looks equivalent and is not: body bytes are copied out as they arrive, so a message waiting for the last two bytes of its body has an empty buffer.


emiago:
I could see point to have this merged, but fix could be simpler. Instead me spending tokens, 
let your AI have 1 regression test, return sip partial error up to stack and handle check in loop instead converting to bool.

fridokus:
Done, all three.

- \`parseStream\` keeps its \`error\` signature and returns \`ErrParseSipPartial\` up instead of folding it into \`nil\`. Net effect is that it is now simpler than before the PR.
- The read loop keeps the last parse result and checks it — no bool.
- One regression test, down from four.

The test splits the message at each of its CRLFs so the LF lands in a read of its own. Without the guard it delivers 0 messages instead of 2, at the first boundary. Full suite passes.

PR #363, opened 2026-09-09
fix: do not hold server transactionStore lock during dial

zuercher:
When serverTxRequest triggers a dial, all other transactions were blocked until the dial completed (by a lock on the transactionStore). If the dial fails due to timeout or is otherwise slow, unrelated transactions can blocked long enough
to cause errors on those transactions.

Fix by using transactionStore's existing get/put and using a singleflight.Group to guarantee only a single transaction is created per key.

Add TestTransactionLayerServerTxCreationDoesNotBlockOtherTransactions which verifies that different transaction 
keys are handled independently.


emiago:
thanks for reporting, but solved without adding sf for now. 
Pls check commit and see does it fixes your problem.

zuercher:
> thanks for reporting, but solved without adding sf for now. Pls check commit and see does it fixes your problem.

Yes, that should resolve the issue. Thanks!

PR #370, opened 2026-09-18
fix: create a server transaction outside the transaction store's lock

fridokus:
`serverTxRequest` held the transaction store's write lock across `serverTxCreate`. Creating a server transaction acquires the connection its responses will use, and when the connection the request arrived on is gone that is a dial to the Via host, which for a peer that went away lasts the whole connect timeout. Every other request in the layer waits on the lock for those ten seconds, times out at its sender, and the senders' now-closed connections make them the next dials. Seen on a proxy carrying ten thousand subscribers: two of four replicas ran that convoy for two hours, 248 goroutines parked on the lock.

The transaction is now created unlocked and inserted under the lock. If a retransmission raced its copy in first, the loser terminates what it created and hands its request to the winner.

Tests: `TestTransactionLayerServerTxCreationDoesNotHoldTheStore`, `TestTransactionLayerRacingCopiesShareOneServerTx`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fridokus:
Duplicate of #366, opened by mistake — same change, same tests. Closing this one.

